PR #18, opened 2020-08-12
Added CMAKE related files

Altaimorin:
This CMAKE build system works for serial, MPI and CUDA versions of installation of QUICK separately.
1.Create a build folder:
`mkdir cmake_build`
`cd cmake_build`

2.Choose either INTEL or GNU as the _compiler_ for build:
Serial version:
`cmake .. -DCOMPILER=<compiler>`
MPI version:
`cmake .. -DMPI=TRUE -DCOMPILER=<compiler>`
CUDA version:
`cmake .. -DCUDA=TRUE -DCOMPILER=<compiler>`

3.Run build command:
`make`
The executables should be found in the bin folder.

-------------------------------------------------
updates on 09/01/2020:
The latest CMAKE system supports building three versions of QUICK separately at the same time and installing executables, libraries(specified as either static or shared) and modules in the correspondingly separate folders. 

Sample commands to build:

`mkdir cmake_build && cd cmake_build`
`cmake .. -DMPI=TRUE -DCUDA_ARCH=volta -DCOMPILER=GNU -DDEBUG=TRUE -DSHARED=TRUE -DNOF=TRUE` (all flags but -DCOMPILER are optional)
`make`
`make install`

To run test, simply do 
`make test`


CMakeLists.txt
```diff
@@ -0,0 +1,75 @@
+cmake_minimum_required(VERSION 3.8.1)
+project(QUICK NONE)
+
+message("before init")
+include(cmake/QuickBuildSystemInit.cmake NO_POLICY_SCOPE)
+message("before enable_language")
+enable_language(Fortran C CXX) #now ENABLED_LANGUAGES is a property but not a variable
```
multiplemonomials, line 7: CUDA should be included in this list, see my comment on FindCUDA.

cmake/CUDAConfig.cmake
```diff
@@ -0,0 +1,32 @@
+if(NOT ${CUDA_ARCH} STREQUAL "")
+	#SET(CUDA_SEPARABLE_COMPILATION ON) # for -dc; no need
+	find_package(CUDA)
```
multiplemonomials, line 3: I think you are getting a bit confused between the FindCUDA module (which was how CMake supported CUDA in versions earlier than 3.10), and the CMake CUDA language (first-class CUDA support for CMake that was added in CMake 3.10).  
`find_package(CUDA)` is part of the former method.  However, to use FindCUDA, you would need to use the `cuda_add_executable` and `cuda_add_library` macros -- you can't use CUDA with standard CMake add_library().  However, your buildscript does add .cu files directly to source lists and build them with add_library().  So, you're actually using the CMake CUDA language, not FindCUDA.

For one thing, this means that your actual minimum CMake version is 3.10, not 3.8.1 like it says in your cmake_minimum_required() call.   Also, it means that there is no need to call `find_package(CUDA)` as long as you add it to the list of languages.

cmake/CUDAConfig.cmake
```diff
@@ -0,0 +1,32 @@
...
+	#message("CUDA_LIBRARIES  is ${CUDA_LIBRARIES}")
+	#message("CUDA_TOOLKIT_ROOT_DIR is ${CUDA_TOOLKIT_ROOT_DIR}")	#this is CUDA_HOME
+	#message("CUDA_cublas_LIBRARY is ${CUDA_cublas_LIBRARY}")	
+	
+	include_directories(${CUDA_INCLUDE_DIRS})	
+
+	add_definitions(-DCUDA)
+	add_definitions(-DDEBUG)
+		
+	if(${CUDA_ARCH} STREQUAL "pascal")
```
multiplemonomials, line 22: This block doesn't properly duplicate the logic from the configure script (see my implementation in CMakeLists.txt).

src/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+endif()
+
+# generate executable quick/quick.MPI/quick.CUDA
+add_executable(${quickexe} main.f90)
+link_directories(${libfolder})
+if("${CUDA_ARCH}" STREQUAL "")
+	message("Not CUDA")
+	target_link_libraries(${quickexe} quicklib blas xc)
+else()
+	target_link_libraries(${quickexe} quicklib gpu_all cublas cusolver xc -lcublas -lcusolver -lcuda -lm -lcudart -lcudadevrt -lstdc++ ${CUDA_LIBRARIES})
```
multiplemonomials, line 42: You don't need all of those `-l` libraries.  CMake will handle linking the CUDA runtime automatically, as well as libstdc++.

src/libxc/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+#	list(APPEND localobjlist "$<TARGET_OBJECTS:${filename}>")
+#endforeach()
+
+foreach(src ${libxcsrc})
+	add_library(${src} OBJECT ${src}.c)
+	list(APPEND localobjlist "$<TARGET_OBJECTS:${src}>")
+endforeach()
+
+
+add_library(libxcf90_la-xc_f OBJECT xc_f.c)
```
multiplemonomials, line 83: Why is this in its own library?

src/libxc/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+
+add_library(libxcf90_la-libxc OBJECT libxc.f90 libxc_funcs.f90)
+list(APPEND localobjlist "$<TARGET_OBJECTS:libxcf90_la-libxc>")
+
+if(NOT ${CUDA_ARCH} STREQUAL "")	#for cuda, add gpu.c, test.cu
+	add_library(gpu OBJECT gpu.c)
+	list(APPEND localobjlist "$<TARGET_OBJECTS:gpu>")
+endif()
+
+add_library(xc SHARED ${localobjlist})
```
multiplemonomials, line 97: Does libxc need to be a shared library?  Since QUICK uses its own customized version of libxc, I feel like installing that version as a shared library risks conflicting with the regular version of libxc if the user installed it.

src/libxc/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+add_library(libxcf90_la-xc_f OBJECT xc_f.c)	
+list(APPEND localobjlist "$<TARGET_OBJECTS:libxcf90_la-xc_f>")
+
+#add_library(libxcf90_la-libxc_funcs OBJECT libxc_funcs.f90)
+#list(APPEND localobjlist "$<TARGET_OBJECTS:libxcf90_la-libxc_funcs>")
+
+add_library(libxcf90_la-libxc OBJECT libxc.f90 libxc_funcs.f90)
+list(APPEND localobjlist "$<TARGET_OBJECTS:libxcf90_la-libxc>")
+
+if(NOT ${CUDA_ARCH} STREQUAL "")	#for cuda, add gpu.c, test.cu
```
multiplemonomials, line 92: I really feel like it would be better to use a separate boolean (`USE_CUDA`?) to control whether we are doing a CUDA build.

src/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+file(GLOB srcsrc "*.f90") # add all *.f90 files
+#list(REMOVE_ITEM ${srcsrc} "${CMAKE_CURRENT_SOURCE_DIR}/quick_api_test.f90" )
+
+foreach(src ${srcsrc})
```
multiplemonomials, line 4: I think that you should be able to combine this logic (and the code in modules/ and subs/) into a single file(GLOB_RECURSE) expression (plus maybe some exclusions and/or REMOVE_ITEMs on the list).

src/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+message("CUDA_ARCH is ${CUDA_ARCH}")
+if(NOT "${CUDA_ARCH}" STREQUAL "")
+    set(quickexe "quick.CUDA")
+	set(test-apiexe "test-api.CUDA")
+endif()
+
+# generate executable quick/quick.MPI/quick.CUDA
+add_executable(${quickexe} main.f90)
+link_directories(${libfolder})
```
multiplemonomials, line 37: You shouldn't need to ever use link_directories() as long as your library targets are set up correctly.  Including a library target name in target_link_libraries() like you did will automatically make cmake link it by the correct full path.  Hooray for less complexity!

src/libxc/maple2c_device/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+lda_c_vwn_2.cu;lda_c_vwn_3.cu;lda_c_vwn_4.cu;lda_c_vwn.cu;\
+lda_c_vwn_rpa.cu;lda_c_wigner.cu;lda_k_tf.cu;lda_k_zlp.cu;\
+lda_x_2d.cu;lda_xc_1d_ehwlrg.cu;lda_xc_ksdt.cu;lda_xc_teter93.cu;\
+lda_x.cu;lda_xc_zlp.cu;lda_x_rel.cu")
+
+
+#file(GLOB maple2c_devicesrc "*.cu")
+foreach(src ${maple2c_devicesrc})
+    get_filename_component(filename ${src} NAME_WE)
+    add_library(${filename}cu OBJECT ${src})
```
multiplemonomials, line 39: I understand why you're doing it (avoiding separable compilation linking issues), but making these each their own library is kinda gross because it clutters up the build output (and makefiles) with a bunch of targets.  I figured out a cleaner solution in my PR: compile all the libxc sources plus gpu_getxc.cu into a single static library.

src/libxc/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
+set(CMAKE_C_FLAGS "-c -I. -O0 -g -fPIC") # ${CCC}
```
multiplemonomials, line 1: It's not great that flags like this are being specified in a bunch of different places.  What if you use a different compiler such as Intel that needs different optimization flags?  It would be a lot better to make constants that are defined to the opt and non-opt flags (for each language) and then activate them when needed.  This is how I've done it in Amber (except on a source file or target basis, not at directory scope).

cmake/CUDAConfig.cmake
Altaimorin: Right below this I have `enable_language(CUDA)`. Shall I also remove it, since I'm going to add `cuda` to `enable_language()` in the main `CMakeLists.txt` per your first comment? Or which way is better(Originally I thought you only need to enable CUDA when you want to build the CUDA version)?

cmake/CUDAConfig.cmake
Altaimorin: I'm a bit confused here. `add_definitions` won't work in this case? What's the difference between `add_definitions`, `compile_options` and`target_compile_options`?

cmake/CUDAConfig.cmake
Altaimorin: If I commented out `find_package(CUDA)` and wherever I enabled CUDA, it always gives me errors like:

_CMakeFiles/quicklib.dir/octree/grid_packer.cpp.o: In function `gpu_get_pfbased_basis_function_lists_new_imp(std::vector<node, std::allocator<node> >*, std::vector<node, std::allocator<node> >*, std::vector<bflist, std::allocator<bflist> >*)':
grid_packer.cpp:(.text+0x5183): undefined reference to `gpu_get_octree_info'
CMakeFiles/quicklib.dir/modules/quick_api_module.f90.o: In function `__quick_api_module_MOD_delete_quick_job':
quick_api_module.f90:(.text+0x210): undefined reference to `gpu_shutdown_'_
...

multiplemonomials: > Right below this I have enable_language(CUDA). Shall I also remove it, since I'm going to add cuda to enable_language() in the main CMakeLists.txt per your first comment?

Oh right I missed that.  Yeah it's easiest to just add it to the enable_languages() call.

multiplemonomials: > If I commented out find_package(CUDA) and wherever I enabled CUDA, it always gives me errors like:

I wonder if there's some conditions that are getting turned off when FindCUDA isn't called...

cmake/CUDAConfig.cmake
multiplemonomials: What I meant is that the `if(${CUDA_ARCH} STREQUAL "pascal")` block doesn't duplicate the logic correctly.  You should also have code to auto-set CUDA_ARCH if the user doesn't provide it, like how the configure script enables different architectures based on CUDA version.

I already implemented this starting [here](https://github.com/multiplemonomials/QUICK/blob/080ab85894ecd3dbe689e7d804d24743142e6d98/CMakeLists.txt#L80), you should be able to just use my implementation (changing some of the variable names).

src/libxc/CMakeLists.txt
Altaimorin: In the old Makefile `xc_f.c` was first compiled into an object file, then the object files were used to build `libxc` library. Probably I don't need the first step in CMAKE?

src/CMakeLists.txt
Altaimorin: I was trying to remove these `-l` libraries but got errors of undefined reference like below:

_[ 98%] Linking Fortran executable ../../bin/test-api.CUDA
cuda/cublas/libcublas.a(fortran_thunking.c.o): In function `cublas_isamax_':
fortran_thunking.c:(.text+0x149): undefined reference to `cublasAlloc'
fortran_thunking.c:(.text+0x195): undefined reference to `cublasSetVector'
fortran_thunking.c:(.text+0x1b9): undefined reference to `cublasFree'
..._

cmake/CUDAConfig.cmake
Altaimorin: This block has been fixed per your suggestion. I just have two naive questions:
1. What's the difference between `CUDA_NVCC_FLAGS` and `CMAKE_CUDA_FLAGS`?
2. For `CUDA_VERSION`~[9.0,10.0) you used `list(APPEND CUDA_NVCC_FLAGS ${SM60FLAGS} ${SM70FLAGS})`. Does that mean for instance `CUDA/9.1.85` would have two sets of `arch` and `code` values?

CMakeLists.txt
Altaimorin: `CUDA` has been added in this list.

cmake/CUDAConfig.cmake
multiplemonomials: CUDA_NVCC_FLAGS is the FindCUDA equivalent of CMAKE_CUDA_FLAGS.  Also, CMAKE_CUDA_FLAGS is space-seperated while CUDA_NVCC_FLAGS is a list.  You should be using CMAKE_CUDA_FLAGS since you're not using FindCUDA.  Also, for the same reason, you should use CMAKE_CUDA_COMPILER_VERSION instead of CUDA_VERSION.  The logic is a lot different between the new and old CUDA libraries.

multiplemonomials: > Does that mean for instance CUDA/9.1.85 would have two sets of arch and code values?

Yep!  This usage causes nvcc to compile the file twice, once for each architecture.

src/CMakeLists.txt
multiplemonomials: Is `cublas` still in the list?  That's what it seems to think is missing...

src/libxc/CMakeLists.txt
multiplemonomials: Yeah I'm pretty sure they can all be in the same library.

src/libxc/CMakeLists.txt
Altaimorin: This has been fixed.

src/libxc/CMakeLists.txt
Altaimorin: `USE_CUDA` has been added to replace `CUDA_ARCH` in the if statements.

src/CMakeLists.txt
Altaimorin: Yes `cublas`is still in the list as the CUDA equivalent for `blas`.

src/libxc/CMakeLists.txt
Altaimorin: This has been fixed.

src/CMakeLists.txt
Altaimorin: I fixed it by `file(GLOB srcsrc ${CMAKE_CURRENT_SOURCE_DIR}/*.f90 ${CMAKE_CURRENT_SOURCE_DIR}/subs/*.f90 ${CMAKE_CURRENT_SOURCE_DIR}/modules/*.f90)`

src/CMakeLists.txt
Altaimorin: This has been commented it out.

src/libxc/CMakeLists.txt
Altaimorin: This line seems to be unnecessary...

Madu86:
@Altaimorin How can I generate quick.rc file when building with cmake?

Madu86:
@Altaimorin Is it possible to make -DCOMPILER options case insensitive? Currently, if you make -DCOMPILER=intel, cmake won't recognize intel compiler.

Madu86:
@vwcruzeiro Chi has rebased cmake build system on v21.03. Can you please test and give your suggestions? You are well aware of the requirements for AMBER integration. I am also testing now.

Madu86:
@Altaimorin How do I specify cudampi and serial builds? The following doesn't seem to work.

`CMake Warning:
  Manually-specified variables were not used by the project:

    CUDAMPI
    SERIAL`

Madu86:
@Altaimorin We need to make sure that following options are working in the cmake build system.

  Available flags are:                                                      
      --prefix <dir> User specified installation directory                  
      --serial       Builds a serial version                                
      --mpi          Compiles MPI parallel version                           
      --cuda         Builds GPU version that utilizes a single NVIDIA GPU   
      --cudampi      Builds multi-GPU version that utilizes multiple NVIDIA GPUs  
      --debug        Compiles debug version                                 
      --debug-time   Compiles a debug version that reports more information
                     on timing
      --shared       Build shared object libraries                          
      --arch <kepler|maxwell|pascal|volta|turing|ampere>                                           
                     Specify gpu architecture. Applicable for cuda and      
                     cudampi versions only. If unspecified, QUICK will be   
                     compiled for several architectures based on the CUDA   
                     toolkit version.                                       
           
      --nof      Disables the compilation of time consuming f functions 
                     in the ERI code of cuda version. Not recommended for 
                     production.
      --amber Install QUICK executables and libaries into AMBER. 
                     Requires specifying AMBER home directory as prefix. 
      --lapack     Use matrix diagonalizer from LAPACK. BLASROOT environment
                     variable must be set with the correct blas installation path. 
                     This path must contain lib and include directories.  
      --mkl          Use matrix diagonalizer from MKL. MKLROOT environment
                     variable must be set with the correct mkl installation path.
                     This path must contain lib and include directories.

  Supported compilers are: gnu, intel, pgi, nvidia

Altaimorin:
> @Altaimorin How do I specify cudampi and serial builds? The following doesn't seem to work.
> 
> `CMake Warning`
> `Manually-specified variables were not used by the project`
> 
> ` CUDAMPI SERIAL`

Serial is the default build type. When CUDA and MPI are selected, CUDA-MPI will be built automatically. I will fix this

Altaimorin:
> @Altaimorin How do I specify cudampi and serial builds? The following doesn't seem to work.
> 
> `CMake Warning`
> `Manually-specified variables were not used by the project`
> 
> ` CUDAMPI SERIAL`

Altaimorin:
> @Altaimorin How do I specify cudampi and serial builds? The following doesn't seem to work.
> 
> `CMake Warning`
> `Manually-specified variables were not used by the project`
> 
> ` CUDAMPI SERIAL`

vwcruzeiro:
@Madu86 @Altaimorin @agoetz : I will be happy to help to test this. How can I compile QUICK specifying the Amber folder as the target to place the libraries and executables? Thank you!

Altaimorin:
> @Madu86 @Altaimorin @agoetz : I will be happy to help to test this. How can I compile QUICK specifying the Amber folder as the target to place the libraries and executables? Thank you!

Let us wait for Jammie to take a look at the CUDAMPI build and fix the issue. We also need him for integration to AMBER.

vwcruzeiro:
I am closing this PR in favor of PR #19

PR #19, opened 2020-08-12
My take on a CMake build system

multiplemonomials:
So I was already about 75% done with implementing a CMake build system when I got Dr. Goetz's email saying that Chi's one was already finished.  Oh well... I figured I'd post what I had anyway so you guys can look it over.  This build system is able to compile and link `quick`, `quick.mpi`, and `quick.cuda`.  However, I haven't tested them yet so I don't know if they run properly.  

I wrote this build system using the library of CMake scripts I developed for Amber.  This library is designed to be used in other programs and you can turn parts of it on and off as needed by including the relevant scripts.  I've already used it to create the standalone build systems for other Amber programs (cpptraj and cphstats).  Yeah it includes a fair amount of extra code files that aren't needed here, but it isn't doing any harm and we can delete them later if needed.  However, this library does provide quite a bit of useful functionality, including:
- Finding MPI and creating targets for it, as well as the adding the `make_mpi_version()` function
- Finding CUDA
- Finding BLAS/MKL and giving an option to switch between them and the internal blas
- Automatically providing correct compile flags for all of Amber's supported compilers

So, it does save quite a lot of work to use this library!  I've gotten into a pretty good groove of using it to hammer out scientific computing build systems as it handles a lot of the hard stuff for you.  Also, I had already wrote a build system for the version of QUICK in the Amber repository, so I was able to draw on that (very old) work to some extent.

Also, this build system is designed to work equally well inside and outside Amber.  It's set up so that it can be directly added as a subdirectory by AmberTools, and it will use the flags and information already present in the amber build system.  If it's not an Amber subdirectory (INSIDE_AMBER = FALSE), then it loads the amber build system library itself.  Importantly, this build system uses Amber's methodology of pan-configuration builds: it doesn't _switch_ to MPI or CUDA versions, it just copies the serial version of the program target and creates a parallel version with MPI or CUDA enabled.   Hopefully this can save you a lot of work when testing the program as you can use one build directory to test all four variants (serial, mpi, cuda, and cudampi).  I've found this method also helps to prevent build system errors to some extent since you have to be more aware of what changes are needed to make the MPI/CUDA version.

Anyway, hope this is useful at least as a reference.  I can explain about it some more at the meeting tomorrow.
Oh also, the pull request for this is showing a bunch of unimportant changes due to me adding the amber cmake-buildscripts, so it's hard to see the actual new code.  I put all the new buildscripts into a separate commit, you can view them [here](https://github.com/merzlab/QUICK/commit/080ab85894ecd3dbe689e7d804d24743142e6d98).


CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+endif()
+
+option(NOF "Disables the compilation of QUICK's time consuming f functions in the ERI code of cuda version. Not recommended for production." FALSE)
+
+# C/Fortran compiler flags
+# --------------------------------------------------------------------
+
+# add -DDEBUG in debug configuration
+set_property(DIRECTORY . PROPERTY COMPILE_DEFINITIONS $<$<CONFIG:Debug>:DEBUG>)
+
```
Altaimorin, line 50: Is there any difference between `.` and `${CMAKE_CURRENT_SOURCE_DIR}`? When I use `GLOB`, it seems that `./*.f90` doesn't work and I have to use `${CMAKE_CURRENT_SOURCE_DIR}/*.f90`

CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+# optimization level
+if(OPTIMIZE)
+	list(APPEND CUDA_NVCC_FLAGS -O3)
+else()
+	list(APPEND CUDA_NVCC_FLAGS -O0)
+endif()
+
+# debug flags
+list(APPEND CUDA_NVCC_FLAGS $<$<CONFIG:Debug>:-g> $<$<CONFIG:Debug>:-G> $<$<CONFIG:Debug>:-DDEBUG>)
+
```
Altaimorin, line 120: Is ${CONFIG} essentially ${CMAKE_BUILD_TYPE}?

src/CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
+# recurse to subdirs
+# --------------------------------------------------------------------
+add_subdirectory(libxc)
```
Altaimorin, line 3: Where do you define ${libxc}?

src/CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+
+foreach(SUBS_SOURCE ${QUICK_SUBS_SOURCES})
+	list(APPEND QUICK_SUBS_SOURCES_FULLPATH "subs/${SUBS_SOURCE}")
+endforeach()
+
+# create serial and MPI libraries
+# --------------------------------------------------------------------
+
+add_library(libquick ${QUICK_MODULES_SOURCES_FULLPATH} ${QUICK_SUBS_SOURCES_FULLPATH} ${QUICK_GENERAL_SOURCES})
+target_compile_options(libquick PRIVATE ${OPT_FFLAGS})
```
Altaimorin, line 79: Where was ${OPT_FFLAGS} defined?

src/CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+config_module_dirs(libquick ${QUICK_SERIAL_MOD_DIR} ${XC_SERIAL_MOD_DIR})
+target_link_libraries(libquick xc octree blas)
+install_libraries(libquick)
+
+# "liblibquick" -> "libquick"
+set_property(TARGET libquick PROPERTY OUTPUT_NAME quick)
+
+if(MPI)
+	make_mpi_version(libquick libquick_mpi LANGUAGES Fortran INSTALL)
```
Altaimorin: This is to make `libquick_mpi` from the existing `libquick`?

src/CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+	set_property(TARGET libquick_mpi PROPERTY OUTPUT_NAME quick_mpi)
+
+	# change link libraries and mod dirs for MPI
+	remove_link_libraries(libquick_mpi xc octree)
+	target_link_libraries(libquick_mpi xc_mpi octree_mpi)
+	config_module_dirs(libquick_mpi ${QUICK_MPI_MOD_DIR} ${XC_MPI_MOD_DIR})
+endif()
+
+if(CUDA)
+	copy_target(libquick libquick_cuda SWAP_SOURCES TO $<TARGET_OBJECTS:cublas_fortran_obj> $<TARGET_OBJECTS:cusolver_obj>)
```
Altaimorin, line 103: Little confused about the function `copy_target`: is `libquick` the ${SOURCE} and `libquick_cuda` the ${DESTINATION}?

src/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+set(QUICK_CUDA_CUDA_SOURCE gpu.cu gpu_type.cu gpu_get2e.cu)
+     
+set(QUICK_CUDA_FORTRAN_SOURCE gpu_write_info.f90)
+
+set_property(SOURCE ${QUICK_CUDA_FORTRAN_SOURCE} PROPERTY COMPILE_OPTIONS ${OPT_FFLAGS})
+
+#sadly FindCUDA doesn't support OBJECT libraries, so we have to use a static cuda_add_library
+cuda_add_library(quick_cuda STATIC ${QUICK_CUDA_FORTRAN_SOURCE} ${QUICK_CUDA_CUDA_SOURCE} 
+	OPTIONS ${CUDA_DEVICE_CODE_FLAGS} -DCUDA)
```
Altaimorin, line 5: I guess the reason why `cuda_add_library` is used here is that you used `find_package(CUDA)` instead of `enable_language(CUDA)`?

src/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+set(QUICK_CUDA_CUDA_SOURCE gpu.cu gpu_type.cu gpu_get2e.cu)
+     
+set(QUICK_CUDA_FORTRAN_SOURCE gpu_write_info.f90)
+
+set_property(SOURCE ${QUICK_CUDA_FORTRAN_SOURCE} PROPERTY COMPILE_OPTIONS ${OPT_FFLAGS})
+
+#sadly FindCUDA doesn't support OBJECT libraries, so we have to use a static cuda_add_library
+cuda_add_library(quick_cuda STATIC ${QUICK_CUDA_FORTRAN_SOURCE} ${QUICK_CUDA_CUDA_SOURCE} 
+	OPTIONS ${CUDA_DEVICE_CODE_FLAGS} -DCUDA) 
+target_compile_options(quick_cuda PRIVATE -DCUDA)
```
Altaimorin: Why do we need `-DCUDA` twice?

src/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+target_link_libraries(quick_cuda xc_gpu xc_cuda)
+make_pic_if_needed(quick_cuda)
```
Altaimorin: What's this function `make_pic_if_needed`?

src/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+if(MPI)
+	cuda_add_library(quick_cuda_mpi STATIC ${QUICK_CUDA_FORTRAN_SOURCE} ${QUICK_CUDA_CUDA_SOURCE} 
+	OPTIONS ${CUDA_DEVICE_CODE_FLAGS} -DMPIV -DCUDA_MPIV) 
+	target_compile_options(quick_cuda_mpi PRIVATE -DMPIV -DCUDA_MPIV)
```
Altaimorin: Shouldn't here be `target_compile_definitions(quick_cuda_mpi PRIVATE -DMPIV -DCUDA_MPIV)`?

src/libxc/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+
+# -------------------------------------------------------------
+# enable optimization
+set_property(SOURCE ${LIBXC_C_SOURCES} PROPERTY COMPILE_FLAGS ${OPT_CFLAGS_SPC})
+set_property(SOURCE ${LIBXC_FORTRAN_SOURCES} PROPERTY COMPILE_FLAGS ${OPT_FFLAGS_SPC})
+
+# -------------------------------------------------------------
+# Generate fortran mangling header
+include(FortranCInterface)
+FortranCInterface_HEADER(${CMAKE_CURRENT_BINARY_DIR}/xc-fortran-mangling.h)
```
Altaimorin, line 125: What is here?

CMakeLists.txt
multiplemonomials: You know, I'm not really sure, I think it kind of depends on the application.  In most cases in CMake paths are interpreted relative to the current source dir -- the only major exception is things like add_custom_command() where you specify a working directory.

CMakeLists.txt
multiplemonomials: Yep.  This is a generator expression that only adds the second part (-g) when the first part is true (configuration is Debug).

src/CMakeLists.txt
multiplemonomials: I don't define it anywhere.  `libxc` is a string literal on this line (keep in mind, CMake will interpret any text as a string literal, even if it's not quoted).

src/CMakeLists.txt
multiplemonomials: Yep.  make_mpi_version() calls copy_target() and links in the MPI libraries.

src/CMakeLists.txt
multiplemonomials: Yep.

src/CMakeLists.txt
Altaimorin: So `libxc` is equal to `${CMAKE_CURRENT_SOURCE_DIR}/libxc` here?

src/cuda/CMakeLists.txt
multiplemonomials: Oh yeah I can get rid of that.  It was because that library contained CUDA code (which only cares about the OPTIONS to cuda_add_library()), and Fortran code (which is affected by target_compile_options()).  But now it's a CUDA only library.

src/cuda/CMakeLists.txt
multiplemonomials: I made it for when you are building a static library but are going to be linking it into a shared library later.  It makes the static library position independent (passes -fPIC) if CMake is set to build shared libraries.  But actually, hmm, it doesn't really apply here because this project is set up to always build shared libraries regardless of BUILD_SHARED_LIBS

src/cuda/CMakeLists.txt
multiplemonomials: Yeah that's more correct I suppose, except you don't use the -D flag there so it becomes
```cmake
target_compile_definitions(quick_cuda_mpi PRIVATE MPIV CUDA_MPIV)
```
The only thing is, FindCUDA is weird about compile options -- it doesn't apply them to .cu sources.  The only options that are actually applied are the ones in the OPTIONS argument above.

src/libxc/CMakeLists.txt
multiplemonomials: I noticed that libxc's config.h is supposed to detect the Fortran mangling (how to name C functions so they're callable from Fortran).  So, I tried to re-add that support (that looks like it was lost when libxc was imported into quick).

src/libxc/config.h
```diff
@@ -22,13 +22,21 @@
...
    case), which must not contain underscores, for linking with Fortran. */
 #define FC_FUNC(name,NAME) name ## _
 
 /* As FC_FUNC, but for C identifiers containing underscores. */
 #define FC_FUNC_(name,NAME) name ## _
 
+#endif
+
 /* The size of a Fortran integer */
 #define FC_INTEGER_SIZE 4
```
Altaimorin, line 42: Is this file called anywhere in CMAKE?

src/CMakeLists.txt
multiplemonomials: yep

cmake/CompilerFlags.cmake
```diff
@@ -0,0 +1,453 @@
...
+		add_flags(C -Wno-unused-variable -Wno-unused-but-set-variable)
+	endif()
+	
+	if(NOT UNINITIALIZED_WARNINGS)
+		add_flags(C -Wno-uninitialized -Wno-maybe-uninitialized)
+	endif()
+	
+	if(${CMAKE_C_COMPILER_VERSION} VERSION_GREATER 4.1)
+		if(SSE)
+			if(TARGET_ARCH STREQUAL x86_64)
```
Altaimorin, line 74: What is `SSE` here?

cmake/CompilerFlags.cmake
```diff
@@ -0,0 +1,453 @@
...
+          		#-mfpmath=sse is default for x86_64, no need to specific it
+          		set(OPT_CFLAGS ${OPT_CFLAGS} "-mtune=native")
+        	else() # i386 needs to be told to use sse prior to using -mfpmath=sse
+          		set(OPT_CFLAGS "${OPT_CFLAGS} -mtune=native -msse -mfpmath=sse")
+         	endif()
+         endif()
+	endif()    
+  
+	if(DRAGONEGG)
+		#check dragonegg
```
Altaimorin, line 84: What's `DRAGONEGG` here?

cmake/CompilerFlags.cmake
```diff
@@ -0,0 +1,453 @@
...
+	if(NOT UNUSED_WARNINGS)
+		add_flags(C -Wno-unused-variable -Wno-unused-but-set-variable)
+	endif()
+	
+	if(NOT UNINITIALIZED_WARNINGS)
+		add_flags(C -Wno-uninitialized -Wno-maybe-uninitialized)
+	endif()
+	
+	if(${CMAKE_C_COMPILER_VERSION} VERSION_GREATER 4.1)
+		if(SSE)
```
Altaimorin, line 73: What's `SSE` here?

cmake/CompilerFlags.cmake
```diff
@@ -0,0 +1,453 @@
...
+	endif()
+endif()
+
+#msvc
+#---------------------------------------------------------------------------------------------------------------------------------------------------------------------
+
+if("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
+	add_flags(C /D_CRT_SECURE_NO_WARNINGS /MP)
+	
+	set(OPT_CFLAGS "/Ox")
```
Altaimorin, line 235: Is `OPT_CFLAGS` a CMAKE built-in variable?

cmake/MPIConfig.cmake
```diff
@@ -0,0 +1,229 @@
...
+			target_compile_definitions(${TARGET} PRIVATE MPI)
+		endif()
+	endmacro()
+
+	# make a MPI version of the thing passed 
+	# also allows switching out sources if needed
+	# INSTALL - causes the new target to get installed in the MPI component to the default location (BINDIR etc)
+	# usage: make_mpi_version(<target> <new name> LANGUAGES <language 1> [<language 2...>] [SWAP_SOURCES <source 1...> TO <replacement source 1...>] INSTALL)
+	function(make_mpi_version TARGET NEW_NAME) 
+
```
Altaimorin, line 84: Can this function be accessed outside this module? I know a macro can do it but not sure about functions.

cmake/MPIConfig.cmake
```diff
@@ -0,0 +1,229 @@
...
+	# make a MPI version of the thing passed 
+	# also allows switching out sources if needed
+	# INSTALL - causes the new target to get installed in the MPI component to the default location (BINDIR etc)
+	# usage: make_mpi_version(<target> <new name> LANGUAGES <language 1> [<language 2...>] [SWAP_SOURCES <source 1...> TO <replacement source 1...>] INSTALL)
+	function(make_mpi_version TARGET NEW_NAME) 
+	
+		# parse arguments
+		# --------------------------------------------------------------------	
+		cmake_parse_arguments(MAKE_MPI "INSTALL" "" "LANGUAGES;SWAP_SOURCES;TO" ${ARGN})
```
Altaimorin, line 87: Shouldn't the first argument of `cmake_parse_arguments` be `MAKE_MPI_VERSION`?

cmake/MPIConfig.cmake
```diff
@@ -0,0 +1,229 @@
...
+				target_link_libraries(${NEW_NAME} mpi_${LANG_LOWERCASE})
+			endif()
+			
+		endforeach()
+		
+		# install if necessary
+		# --------------------------------------------------------------------
+		if(MAKE_MPI_INSTALL)
+			if("${TARGET_TYPE}" STREQUAL "EXECUTABLE")
+				install(TARGETS ${NEW_NAME} DESTINATION ${BINDIR} COMPONENT MPI)
```
Altaimorin, line 147: I'm a little confused about `COMPONENT MPI`. Could you explain it a bit more?

cmake/CompilerFlags.cmake
multiplemonomials: That's an option in CompilationOptions.  For certain compilers (Intel?) it's used to select SSE optimization flags.

cmake/CompilerFlags.cmake
multiplemonomials: OPT_CFLAGS is a variable containing optimized C flags, it's created by the amber buildscripts.

cmake/MPIConfig.cmake
multiplemonomials: Yep it can be accessed anywhere.

cmake/MPIConfig.cmake
multiplemonomials: Nope -- that first argument is a prefix which is applied to the variables that [cmake_parse_arguments()](https://cmake.org/cmake/help/latest/command/cmake_parse_arguments.html) returns.  Yeah that function is one of the weirdest parts of CMake.

cmake/MPIConfig.cmake
multiplemonomials: That's used for Amber's install components, it doesn't really matter here, but it isn't doing any harm.  The main use of it is to let you split a program into different components for packaging.

cmake/CopyTarget.cmake
```diff
@@ -0,0 +1,145 @@
...
+# Allows switching out sources if needed.
+# usage: copy_target(<target> <new name> [SWAP_SOURCES <source 1...> TO <replacement source 1...>])
+
+function(copy_target SOURCE DESTINATION)
+
+	#message("Copying target ${SOURCE} -> ${DESTINATION} --------------------------------------")
+	# parse arguments
+	# --------------------------------------------------------------------	
+	cmake_parse_arguments(COPY_TARGET "" "" "LANGUAGES;SWAP_SOURCES;TO" ${ARGN})
+
```
Altaimorin, line 25: In the call `copy_target(libquick libquick_cuda SWAP_SOURCES TO $<TARGET_OBJECTS:cublas_fortran_obj> $<TARGET_OBJECTS:cusolver_obj> ${QUICK_CUDA_FORTRAN_SOURCE})`,
do `$<TARGET_OBJECTS:cublas_fortran_obj> $<TARGET_OBJECTS:cusolver_obj> ${QUICK_CUDA_FORTRAN_SOURCE}` all become ${COPY_TARGET_TO}, while ${COPY_TARGET_SWAP_SOURCES} and ${COPY_TARGET_LANGUAGES} are simply empty strings?

cmake/InstallDirs.cmake
```diff
@@ -0,0 +1,43 @@
...
+	
+	if("${CMAKE_INSTALL_POSTFIX}" MATCHES "^/")
+		string(REGEX REPLACE "^/" "" CMAKE_INSTALL_POSTFIX "${CMAKE_INSTALL_POSTFIX}")
+	endif()
+endif()
+
+# install subdirectory setup
+# --------------------------------------------------------------------	
+
+set(BINDIR "${CMAKE_INSTALL_POSTFIX}bin") #binary subdirectory in install location
```
Altaimorin, line 31: A little confuse here: Why do you use `${CMAKE_INSTALL_POSTFIX}` but not `${CMAKE_INSTALL_PREFIX}`? I guess `${CMAKE_INSTALL_POSTFIX}` is not an absolute path?

CMakeLists.txt
```diff
@@ -158,6 +154,23 @@ configure_package_config_file(QUICKConfig.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/Q
...
+# Additional files
+
+# install modules
+if(NOT INSIDE_AMBER)
+	install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/amber-modules DESTINATION include)
+endif()
+
+# install data in basis dir
+install(DIRECTORY basis DESTINATION .)
+
```
Altaimorin, line 169: Aren't `basis` and `test` already existent?

multiplemonomials: Yes, but these need to be copied to the install location so that tests can be run there.

cmake/InstallDirs.cmake
multiplemonomials: Oh, this is another Amber-specific thing that doesn't apply to QUICK; it's to do with creating certain types of .tar packages correctly.

CMakeLists.txt
Altaimorin: Sorry but aren't they already in the current directory (`.`)? Why to set the `DESTINATION` to `.` again?

cmake/CopyTarget.cmake
multiplemonomials: Yep, you got it.

CMakeLists.txt
multiplemonomials: The DIRECTORY argument to install() doesn't work with empty strings IIRC.  So you have to give it some kind of file path to work with.

Altaimorin: I must misunderstand something here. I think `basis` and `test` are already in the current directory as source codes before running CMAKE, while the `install` command here seems to copy them from current directory to current directory. Please correct me.

CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+# extra CUDA flags
+list(APPEND CUDA_NVCC_FLAGS -Xptxas=-v -use_fast_math)
+
+if(SHARED AND (TARGET_LINUX OR TARGET_OSX))
+	list(APPEND CUDA_NVCC_FLAGS --compiler-options -fPIC)
+endif()
+
+# SPDF
+if(NOT NOF)
+	list(APPEND CUDA_NVCC_FLAGS -DCUDA_SPDF)
```
Altaimorin, line 128: Is `CUDA_SPDF` a macro preprocessor? So essentially macro preprocessors can be passed by either compilation flags or definitions?

multiplemonomials: Yes to both of those.

CMakeLists.txt
Altaimorin: Is `CONFIG` an option only passed by the command line, like `cmake .. -DCOMPILER=GNU -DCUDA=TRUE -DCMAKE_INSTALL_PREFIX=$(pwd)/../install -DCONFIG=Debug`, or specified somewhere else?

CMakeLists.txt
Altaimorin: It seems that `.` here stands for the `./install` directory but not the current directory?

Altaimorin: Is `.` ${CMAKE_INSTALL_PREFIX} ?

cmake/ConfigModuleDirs.cmake
```diff
@@ -0,0 +1,39 @@
+#Function which (a) sets the module output directory for a target, and (b) includes other module directories
+#preserves the include directories already set on the target, adding them at the end of the list.
+
+# It removes all previous include directories pointing to Amber's modules folder,
+# so that if you use copy_target() you can set new module paths without leaving the old ones lingering
+# at the end of the include path like a bad smell.
+
+function(config_module_dirs TARGETNAME TARGET_MODULE_DIR) #3rd optional argument: extra module include directories
+
```
Altaimorin, line 11: In the call `config_module_dirs(libquick quick/serial libxc/serial)`, `libquick` is ${TARGETNAME}, are both `quick/serial` and `libxc/serial` ${TARGET_MODULE_DIR} ? If so, how do CMAKE do the split?

cmake/CopyTarget.cmake
```diff
@@ -0,0 +1,145 @@
+# Provides the copy_target() function, which is used to make duplicate MPI and OpenMP versions of targets.
+
+# copies the value of a property from one target to another.  Does nothing if the property is not defined on the SOURCE.
+function(copy_property PROPERTY SOURCE DESTINATION)
+	get_property(PROP_IS_DEFINED TARGET ${SOURCE} PROPERTY ${PROPERTY} SET)
```
Altaimorin, line 5: Since you put `SET` here, does that mean in the future calls the copied property is alway a boolean?

.github/workflows/serial.yml
```diff
@@ -17,3 +17,22 @@ jobs:
...
       - name: Test serial
         run: make test
+  cmake-build:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install deps
+        run: sudo apt-get -y install gfortran cmake
+      - name: Configure serial
+        run: |
```
Altaimorin, line 27: What does a single pipe do here?

multiplemonomials: I think it's used as a line continuation.

cmake/CopyTarget.cmake
multiplemonomials: No, this gets if the property ${PROPERTY} is defined for the current target.

cmake/ConfigModuleDirs.cmake
multiplemonomials: These are just file paths under the amber_modules directory, they don't reference target names at all.

CMakeLists.txt
multiplemonomials: yep

cmake/CopyTarget.cmake
Altaimorin: In the documentation it says:
`If the SET option is given the variable is set to a boolean value indicating whether the property has been set.` What `variable` is being talked about here? If it's not ${PROP_IS_DEFINED}, what actually does `SET` do?

cmake/AmberBuildSystemInit.cmake
```diff
@@ -0,0 +1,73 @@
...
+
+if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
+	message(FATAL_ERROR "You are building in the source directory.  ${PROJECT_NAME} does not support this, since it would obliterate the Makefile build system.")
+endif()
+
+# includes
+# --------------------------------------------------------------------
+
+#Basic utilities.  These files CANNOT use any sort of compile checks or system introspection because no languages are enabled yet
+include(CMakeParseArguments)
```
Altaimorin, line 55: Why do you include `CMakeParseArguments`? Isn't it a built-in function in CMAKE?

cmake/CudaConfig.cmake
```diff
@@ -0,0 +1,113 @@
...
+# With CMake 3.7, FindCUDA.cmake crashes when crosscompiling.
+
+if(CROSSCOMPILE)
+	message(STATUS "CUDA disabled when crosscompiling.")
+	set(CUDA FALSE)
+else()
+
+	# first, find CUDA.
+	find_package(CUDA)
+	option(CUDA "Build ${PROJECT_NAME} with CUDA GPU acceleration support." FALSE)
```
Altaimorin, line 12: Since `CUDA` is passed from the command line `-DCUDA=TRUE`, why do we still set a boolean variable for it?

CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+endif()
+if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "Intel")
+	set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -cpp -diag-disable 8291")
+	set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} -traceback")
+endif()
+
+# CUDA compiler flags
+# --------------------------------------------------------------------
+
+# cancel Amber arch flags, because quick only supports SM6.0 and lower
```
Altaimorin: Do we already have the following cuda settings in `CudaConfig.cmake`?

multiplemonomials: Yeah... Amber's needed cuda flags are different from QUICK's, so I just override the amber flags in the top level CMakeLists.txt.

cmake/AmberBuildSystemInit.cmake
multiplemonomials: Hmm, apparently it isn't needed anymore, but it [used to be](https://cmake.org/cmake/help/v3.6/module/CMakeParseArguments.html) when Amber targeted CMake 3.1.

cmake/CudaConfig.cmake
multiplemonomials: The option is there to create a cache variable which can be set on the command line, or using the config GUI.  Basically, the option() call handles the case where you _don't_ pass `-DCUDA` on the command line, setting a false default.

src/libxc/maple2c_device/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+	lda_c_lp96.cu lda_c_ml1.cu lda_c_pw.cu 
+	lda_c_pz.cu lda_c_rc04.cu lda_c_rpa.cu lda_c_vwn_1.cu 
+	lda_c_vwn_2.cu lda_c_vwn_3.cu lda_c_vwn_4.cu lda_c_vwn.cu 
+	lda_c_vwn_rpa.cu lda_c_wigner.cu lda_k_tf.cu lda_k_zlp.cu 
+	lda_x_2d.cu lda_xc_1d_ehwlrg.cu lda_xc_ksdt.cu lda_xc_teter93.cu 
+	lda_x.cu lda_xc_zlp.cu lda_x_rel.cu
+
+	# gpu_getxc.cu references functions inside libxc_cuda, so with how CMake separable
+	# compilation linking works it's a lot simpler to add it as a source file here.
+	../../cuda/gpu_getxc.cu)
```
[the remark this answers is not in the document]
multiplemonomials: I definitely ran into this at one point... I think it's caused by it trying to build for an incorrect (too old) cuda architecture.

src/CMakeLists.txt
Altaimorin: Trivial question: If `MPI` is passed by the command line `-DMPI=TRUE`, what's the difference between `if(MPI)` and `if(${MPI})`? Seems both work?

src/CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+set_property(TARGET libquick PROPERTY OUTPUT_NAME quick)
+
+if(MPI)
+	make_mpi_version(libquick libquick_mpi LANGUAGES Fortran INSTALL)
+	target_compile_definitions(libquick_mpi PRIVATE MPIV)
+	set_property(TARGET libquick_mpi PROPERTY OUTPUT_NAME quick_mpi)
+
+	# change link libraries and mod dirs for MPI
+	remove_link_libraries(libquick_mpi xc octree)
+	target_link_libraries(libquick_mpi xc_mpi octree_mpi)
```
Altaimorin: In `target_link_libraries(libquick_mpi xc_mpi octree_mpi)`, is there a possibility that the library order matters?

src/CMakeLists.txt
multiplemonomials: Yes they are essentially the same.

src/CMakeLists.txt
multiplemonomials: The library that appears first is significant, the order of the other ones doesn't matter (except in esoteric cases with static libraries).

cmake/ConfigModuleDirs.cmake
Altaimorin: So what is the variable that `libquick` corresponds to in this call?

Altaimorin: I feel that this function is important if I want to build different versions of `QUICK` at the same time.

multiplemonomials: libquick corresponds to TARGETNAME, yeah.  It's the first argument in the function call, so it goes to the first argument in the list.

src/libxc/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+
+# -------------------------------------------------------------
+# create libraries
+
+add_library(xc STATIC ${LIBXC_C_SOURCES} ${LIBXC_FORTRAN_SOURCES})
+make_pic_if_needed(xc)
+target_include_directories(xc PUBLIC . ${CMAKE_CURRENT_BINARY_DIR})
```
Altaimorin, line 132: Isn't ${CMAKE_CURRENT_SOURCE_DIR} automatically included? By the way, why to include ${CMAKE_CURRENT_BINARY_DIR} as well?

CMakeLists.txt
```diff
@@ -0,0 +1,147 @@
...
+
+
+if("${CMAKE_C_COMPILER_ID}" STREQUAL "Intel")
+	set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -traceback")
+endif()
+
+
+if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "GNU")
+	set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -cpp")
+	add_definitions(-DGNU)
```
Altaimorin, line 55: If `target_compile_definitions` was used here, `-DGNU` would be `GNU`. But if `target_compile_options` was used then nothing changes. Is that correct?

src/libxc/CMakeLists.txt
multiplemonomials: No, no directories are automatically included by default.

CMakeLists.txt
multiplemonomials: yep!

src/CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+endforeach()
+
+foreach(SUBS_SOURCE ${QUICK_SUBS_SOURCES})
+	list(APPEND QUICK_SUBS_SOURCES_FULLPATH "subs/${SUBS_SOURCE}")
+endforeach()
+
+# create serial and MPI libraries
+# --------------------------------------------------------------------
+
+add_library(libquick ${QUICK_MODULES_SOURCES_FULLPATH} ${QUICK_SUBS_SOURCES_FULLPATH} ${QUICK_GENERAL_SOURCES})
```
Altaimorin: I got trouble building `libquick_mpi_cuda` as a `SHARED` library here using `copy_target` like what you did below, while `STATIC` worked fine. To my surprise `libquick_mpi_cuda` was built before `libquick_cuda` and `libquick_mpi`. Is that where the problem arose? Should I use `add_dependencies` after calling `copy_target`?

src/libxc/maple2c_device/CMakeLists.txt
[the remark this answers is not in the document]
Altaimorin: Do you know how to fix it?

src/cuda/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+endif()
+
+# CUBLAS fortran thunking
+#------------------------------------------------------------------------------------------
+
+set(CUBLAS_FORTRAN_SOURCE cublas/fortran_thunking.c)
+
+add_library(cublas_fortran_obj OBJECT ${CUBLAS_FORTRAN_SOURCE})
+set_property(TARGET cublas_fortran_obj PROPERTY COMPILE_OPTIONS ${OPT_CFLAGS}) 
+set_property(TARGET cublas_fortran_obj PROPERTY COMPILE_DEFINITIONS CUBLAS_USE_THUNKING)
```
Altaimorin, line 24: Seems that `set_property` doesn't allow me to set the scope for `COMPILE_DEFINITIONS`. Is the default scope private?

src/CMakeLists.txt
multiplemonomials: I don't know, I think I'd have to see your code.  copy_target() is definitely supposed to have the copy build after the original.

src/cuda/CMakeLists.txt
multiplemonomials: You can only use the PRIVATE and PUBLIC keywords with `target_compile_definitions()`, `target_compile_options()`, etc.  It doesn't work with set_property().  Internally, those keywords just control what properties are actually set:
- With PUBLIC,  `target_compile_definitions()` sets properties `INTERFACE_COMPILE_DEFINITIONS` and `COMPILE_DEFINITIONS`.
- With PRIVATE,  `target_compile_definitions()` sets `COMPILE_DEFINITIONS` only.
- With INTERFACE,  `target_compile_definitions()` sets `INTERFACE_COMPILE_DEFINITIONS` only.

multiplemonomials:
Okay, I think this should be about ready to go.  It compiles fine with CUDA, MPI, and serial on my Linux GCC machine.  It definitely needs to be tested on more environments though.

multiplemonomials:
The last thing I need to do is add MIRP, I'll take a look at that in the next couple days.

PR #22, opened 2020-09-24
QUICK library installation into amber lib directory

Madu86:
This pull request contains modifications to copy quick libraries into $AMBERHOME/lib folder. Running `make aminstall` will now copy the following libraries. 
Serial version:
    libblas-quick.so  libquick.so  libxc.so

mpi version:
    libblas-quick.so  libquick-mpi.so  libxc.so

cuda version:
    libquick-cuda.so  libxc-cuda.so

cudampi version:
    libquick-cudampi.so  libxc-cuda.so

Running `make amuninstall` will remove them from $AMBERHOME/lib folder.


vwcruzeiro:
@Madu86 : I see some changes in the `basis` folder (some delitions and some additions). Was that intentional?

vwcruzeiro (changes requested):
Currently, the libraries are being copied to the $AMBERHOME/lib folder. We need to do symbolic links instead, pointing to the original libraries inside Quick's build folder.

No need to copy the BLAS library to $AMBERHOME/lib.

merzlab:
> Currently, the libraries are being copied to the $AMBERHOME/lib folder. We need to do symbolic links instead, pointing to the original libraries inside Quick's build folder.
> 
> No need to copy the BLAS library to $AMBERHOME/lib.

I tried the symlink approach. First, this creates a problem with recompilation since you are not allowed to overwrite the existing symlinks (at least on the platform I was testing).  Second, if we decide to extend QUICK for windows platform in future, we dont have to change this again. If copying doesnt create a problem for AMBER installation, I prefer to keep it this way.

Madu86:
> Currently, the libraries are being copied to the $AMBERHOME/lib folder. We need to do symbolic links instead, pointing to the original libraries inside Quick's build folder.
> 
> No need to copy the BLAS library to $AMBERHOME/lib.

I tried the symlink approach. First, this creates a problem with recompilation since you are not allowed to overwrite the existing symlinks (at least on the platform I was testing).  Second, if we decide to extend QUICK for windows platform in future, we dont have to change this again. If copying doesnt create a problem for AMBER installation, I prefer to keep it this way.

Madu86:
> @Madu86 : I see some changes in the `basis` folder (some delitions and some additions). Was that intentional?

Yes, please don't worry about this.

vwcruzeiro (approved):

PR #24, opened 2020-10-02
Updates for AMBER compatible install and uninstall

Madu86:
I have made the following changes to configure script. 
1) Removed --amprefix and introduced --amber flag. 

This new flag requires the user to specify AMBER_HOME as --prefix. So one can configure for AMBER compatible QUICK installation as follows.
  `./configure --serial --mpi --cuda --cudampi --arch volta --shared --amber --prefix $AMBER_HOME gnu`

The script will verify the $AMBER_HOME by checking if AmberTools directory exist inside the given prefix.

2) Modified Makefiles to eliminate aminstall, amunistall dependency 
After the configuration, the user should run `make` and then `make install` to install executables  and libraries into bin and lib folders in $AMBER_HOME. We will check for the existence of bin and lib directories during `make install`. Uninstallation is done by running `make uninstall`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,24 +73,24 @@ install: $(INSTALLTYPES)
 	@echo  "Please add the following into your .bash_profile or .bashrc file."
 	@echo  "      export QUICK_BASIS=$(installfolder)/basis"
 
-noinstall:
+noinstall: all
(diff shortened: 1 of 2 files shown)


Madu86:
I have also updated the main Makefile such that one can run `make install` without running `make` first.

PR #26, opened 2020-10-28
Improvements to XC and one electron codes

Madu86:
This pull request contains following improvements to XC, one electron integral and dipole codes. 
1) grid point information repacking scheme for better scalability of XC calculations in multi-GPU version
2) integer atomic ops in the XC gradient kernels 
3) one electron integral prescreening
4) asynchronous ERI kernel launching to hide OEI gradient times
5) bug fix in the dipole code


agoetz (approved):
Looks all good to me

PR #27, opened 2020-11-02
Miscellaneous improvements 

Madu86:
This pull request contains following improvements.
1) Correct timing: bug fixes, printing cumulative times in quick output from amber runs
2) New test cases, dacdiff compatible reference quick output files
3) Improved printing: removed printing radius of significance into output by default 
4) Verbose build option


vwcruzeiro:
Hello @Madu86 . What is the procedure to run the tests now?

vwcruzeiro:
Please let me know about the verbose changes as well. How to compile with or without verbose? Thanks!

Madu86:
> Hello @Madu86 . What is the procedure to run the tests now?

The procedure remains same for running standard quick tests (i.e. `make test`). For running tests from amber side and compare using dacdiff, you might want to make changes in the amber scripts. At the end of installation, quick executables will be inside $AMBER_HOME/bin. The test cases are in $QUICK_HOME/tests and saved output files are inside $QUICK_HOME/tests/save. In these saved output files, lines that shouldn't be compared have been marked with '|', consistently with your example.

Madu86:
> Please let me know about the verbose changes as well. How to compile with or without verbose? Thanks!

By default, we will have non-verbose build. To enable verbose build, you should configure with --verbose flag. For example, to configure a verbose serial build:
`./configure --serial --verbose intel`

agoetz:
> > Please let me know about the verbose changes as well. How to compile with or without verbose? Thanks!
> 
> By default, we will have non-verbose build. To enable verbose build, you should configure with --verbose flag. For example, to configure a verbose serial build:
> `./configure --serial --verbose intel`

Not important, but you could use a variable `verbose`, set it to an empty string by default, and test for its value. Then the user can either do a `make` to get the neat non-verbose output, or `make verbose=True` to get verbose output, without having to reconfigure.

agoetz (approved):
Looks great.

agoetz:
Tests pass on chinotto with gnu 8.3.1. Vinicius also tested. I am going to merge this.

PR #28, opened 2020-11-08
Latest updates for output generation

Madu86:
This pull request contains following updates.
1) Write statements have been updated with pipe symbols such that QUICK output files are comparable using AMBER dacdiff tool. 
2) Unnecessary output from sad guess procedure has been eliminated.  
3) Information from octree algorithm such as the number of grid points before and after pruning, number of significant basis and primitive functions for XC calculations are printed into the output.


agoetz (changes requested):
Why do we want to use dacdif for QUICK tests? The current test system seems to work quite well. We can use dacdif for Amber mdout, mdcrd, restrt files from QM/MM runs. Shouldn't we be able to call the QUICK testing infrastructure from Amber's testing infrastructure?

If we really want to use dacdif, then we should prefix *only* lines that are expected to have different output depending on installation / execution environment (e.g. time stamps, file paths, number of MPI tasks, GPU or other hardware info, timing information, or output that may depend on underlying precision of hardware or compiler). Everything else should be identical between different runs (e.g. job settings, number of basis functions, and all convergence data) and must pass a diff (also dacdif).

I would rather postprocess QUICK outputs via e.g. awk and then run dacdif on those. The reason is that different parts of the output may have different requirements to tested accuracy. E.g. nuclear repulsion energy to 12 decimal places, SCF energy to 6 decimal places, gradients to 4 decimal places.

vwcruzeiro:
> Why do we want to use dacdif for QUICK tests? The current test system seems to work quite well. We can use dacdif for Amber mdout, mdcrd, restrt files from QM/MM runs. Shouldn't we be able to call the QUICK testing infrastructure from Amber's testing infrastructure?
> 
> If we really want to use dacdif, then we should prefix _only_ lines that are expected to have different output depending on installation / execution environment (e.g. time stamps, file paths, number of MPI tasks, GPU or other hardware info, timing information, or output that may depend on underlying precision of hardware or compiler). Everything else should be identical between different runs (e.g. job settings, number of basis functions, and all convergence data) and must pass a diff (also dacdif).
> 
> I would rather postprocess QUICK outputs via e.g. awk and potentially create multiple files (e.g. energies, gradients) and then run dacdif on those (compare to energies.save, gradients.save). The reason is that different parts of the output may have different requirements to tested accuracy. E.g. nuclear repulsion energy to 12 decimal places, SCF energy to 6 decimal places, gradients to 4 decimal places. But I don't see a need to use dacdif.

@agoetz , I had a long discussion with @Madu86 about this, where all the options have been evaluated. This is what @Madu86 decided: we will not touch Quick's test suite, and we will create a separate test script compatible with Amber instead.

agoetz:
> @agoetz , I had a long discussion with @Madu86 about this, where all the options have been evaluated. This is what @Madu86 decided: we will not touch Quick's test suite, and we will create a separate test script compatible with Amber instead.
OK, sounds good, thanks for clarifying @vwcruzeiro . @Madu86 in this case we do not need bars in most places of the output. It looks odd e.g. to prefix the nuclear repulsion with a bar but not the total energy. I will leave this PR open for now.

test/Run.tests.amber
```diff
@@ -0,0 +1,28 @@
...
+   export DO_PARALLEL=" "
+fi
+
+if [ -z "$QUICK_BASIS" ]; then
+  export QUICK_BASIS="../basis"
+fi
+
+for i in {ene,opt}*.in
+do
+  $DO_PARALLEL $TESTsander $i < /dev/null || error
```
agoetz, line 23: How does this work? Is there an error function defined?

The tests that I wrote for Amber are derived from sander tests that use csh, which has completely different syntax but provides a goto function:
```
#!/bin/csh -f
#TEST-PROGRAM sander
...
$DO_PARALLEL $TESTsander -O -c $inpcrd -o $mdout < dummy || goto error
...
/bin/rm -f restrt mdin mdinfo dummy
exit(0)

error:
echo "  ${0}:  Program error"
exit(1)
```
This then prints the name of the test script and Program error, and returns 1 if the program fails for some reason.

agoetz (approved):
Looks good. Please go ahead and merge if the tests work in Amber. Thank you!

test/Run.tests.amber
vwcruzeiro: `error` is defined in the following line:

> . ../../../../test/program_error.sh

This script will only work when executed from $AMBERHOME/AmberTools/src/quick/test

vwcruzeiro:
> Looks good. Please go ahead and merge if the tests work in Amber. Thank you!

@agoetz , we should not merge this PR request yet. I found out that serial CPU and parallel CPU codes were giving diverging outputs. Here are a few examples:

```
147c147
<  ALPHA ELECTRON DENSITY    =   22.0000152902
>  ALPHA ELECTRON DENSITY    =    0.0469886283
148c148
<  BETA ELECTRON DENSITY     =   22.0000152902
>  BETA ELECTRON DENSITY     =    0.0469886283
```

```
 @ Begin MP2 Calculation                                         @ Begin MP2 Calculation
Current memory usage=0.000000E+00M                            | CURRENT MEMORY USAGE=0.149012E-03M
TOTAL STEP          =     0                                   | TOTAL STEP          =     1
Second order energy=     0.000000000                          | EFFECT INTEGRALS    =   34780
quick_qm_struct%EMP2               =  -152.050824032          | SECOND ORDER ENERGY =    -0.403468374
                                                              > EMP2                =  -152.454292406
 @ End MP2 Calculation                                           @ End MP2 Calculation
```

I suspect these issues were never identified before because only energies and forces are checked in the main Quick test suite.

@Altaimorin already fixed some of these issues, and @Madu86 will take care of the rest whenever he has a chance.

I am now going to test `quick.cuda.MPI` now to check if all is good with this executable.

test/Run.tests.amber
agoetz: This makes sense. Thanks for explaining!

agoetz:
Good catch @vwcruzeiro.

Indeed, I do not remember noticing that there is a bug in computing the electron density in the MPI code. It is a good idea to test any properties that are computed (or can be computed, e.g. also dipole moment).

Regarding MP2, last I remember it did not work with MPI. Madu's test script switches to serial execution by requesting a single processor for MP2 MPI tests, although the tests should not be run at all if there is no working MPI version. Of course, if @Altaimorin can fix the MPI implementation, that would be best.

PR #29, opened 2020-11-09
Improvements for DIIS

Madu86:
This pull request contains an improvement to diis of cuda and cudampi versions. The formation of the density matrix becomes expensive as the number of basis functions goes up so I have ported the corresponding piece to GPU. The observed times are as follows (left and right are old and new respectively).
1) water B3LYP/cc-pvDZ, 25 basis functions: No significant improvement 
      TOTAL DII TIME      =     0.567264000(  8.20%)	      |	      TOTAL DII TIME      =     0.560116000(  8.03%)
       TOTAL DIAG TIME    =     0.309894000(  4.48%)     |	            TOTAL DIAG TIME    =     0.308439000(  4.42%)

2) water30 B3LYP/def2-svp, 750 basis functions: Small improvement
      TOTAL DII TIME      =     4.267836000(  7.94%)	      |	      TOTAL DII TIME      =     3.670340000(  6.91%)
            TOTAL DIAG TIME    =     0.852865000(  1.59%)     |	            TOTAL DIAG TIME    =     0.847015000(  1.59%)

3) Taxol B3LYP/6-31G**, 1160 basis functions: Small improvement (see here for full comparison: https://pastebin.com/vjRvHvtA)
      TOTAL DII TIME      =    17.012937000(  4.39%)	      |	      TOTAL DII TIME      =    13.589752000(  3.54%)
            TOTAL DIAG TIME    =     1.646698000(  0.42%)     |	            TOTAL DIAG TIME    =     1.649029000(  0.43%)

4) Olestra B3LYP/6-31G**, 4015 basis functions: ~60% improvement (see here for full comparison: https://pastebin.com/Mfp5FB8f)
      TOTAL DII TIME      =   570.893714000( 39.06%)	      |	      TOTAL DII TIME      =   197.258320000( 18.17%)
            TOTAL DIAG TIME    =    14.717386000(  1.01%)     |	            TOTAL DIAG TIME    =    14.802130000(  1.36%)

The new implementation is better for the cuda and cudampi versions. In the olestra example, the formation of density matrix during a given SCF iteration takes about 21 s using the old implementation. This becomes ~0.5 s using the new.


src/cuda/gpu_custom_matops.cu
```diff
@@ -0,0 +1,81 @@
...
+  int totalThreads = blockDim.x*gridDim.x;
+
+  for (QUICKULL gid = offset; gid < devSim_matop.nbasis * devSim_matop.nbasis; gid += totalThreads) {
+
+    QUICKULL i = (QUICKULL) (gid/devSim_matop.nbasis);
+    QUICKULL j = (QUICKULL) (gid - i*devSim_matop.nbasis);
+
+    QUICKDouble denseji=0.0;
+
+    for(int k=0; k<devSim_matop.nElec/2; k++){
```
agoetz, line 74: This will only work for a RHF/RKS density matrix.

src/cuda/gpu_custom_matops.cu
```diff
@@ -0,0 +1,81 @@
...
+      denseji = denseji + LOC2(devSim_matop.co, j, k, devSim_matop.nbasis, devSim_matop.nbasis) * 
+                          LOC2(devSim_matop.co, i, k, devSim_matop.nbasis, devSim_matop.nbasis);
+    } 
+
+    LOC2(devSim_matop.dense, j, i, devSim_matop.nbasis, devSim_matop.nbasis) = denseji * 2.0;
```
agoetz, line 79: Same here, factor 2.0 only valid for a closed shell RHF/RKS calculation.

src/cuda/gpu_custom_matops.cu
```diff
@@ -0,0 +1,81 @@
...
+    totTime+=time;
+    fprintf(gpu->debugFile,"Time to form new density matrix:%f ms total time:%f ms\n", time, totTime);
+    cudaEventDestroy(start);
+    cudaEventDestroy(end);
+#endif
+
+}
+
+// compute new density matrix
+__global__ void get_dmx_kernel(){
```
agoetz, line 62: This implementation is inefficient because it does not use cache blocking. It would be better to use DGEMM (from cuBLAS on GPU and BLAS on CPU).

src/scf.f90
```diff
@@ -528,6 +528,10 @@ subroutine electdiis(jscf)
 
          call CopyDMat(quick_qm_struct%dense,quick_scratch%hold,nbasis) ! Save DENSE to HOLD
 
+#if defined(CUDA) || defined(CUDA_MPIV)
+         call gpu_get_dmx(quick_qm_struct%co, quick_qm_struct%dense)
+#else
+
          ! PIJ=SIGMA[i=1,nelec/2]2CJK*CIK
          do I=1,nbasis
```
agoetz, line 536: This old code to compute the density matrix is very inefficient. No cache-blocking is employed. It is limited by memory bandwidth.

agoetz (commented):
I am leaving a few comments here regarding performance. The code is correct but not efficient. I can see why computing the density matrix becomes expensive, because it is a matrix multiplication and scales as O(N^3). I hope I get this all right off the top of my head: The one-particle density matrix is given as

P_{mn} = sum_i f_i * C_{mi} * C{ni}, where m,n are basis functions, f_i are occupation numbers, and i are MOs. If occupation numbers are integers, such as in RHF and UHF, then we can write this is matrix notation as

P = f * C * C^T, where C is the (rectangular) matrix of occupied MO coefficients and f is 1 for alpha/beta densities in UHF, or 2 in RHF.

Both for the CPU and the GPU code it is important to employ cache blocking for matrix multiplication, i.e. to to read only sub-blocks into cache and do the matrix multiply on the sub-blocks. Otherwise you have to fetch data multiple times from RAM or global memory, which is slow. Already for water30 you have 300 electrons and 750 basis functions with an SVP basis, which means almost 1MB for occupied MO coefficients. For water60 you are at 3.5MB. L1 cache usually is something around 32kB. On a GPU you would want to work with shared memory.

The best would thus be to use DGEMM routines from BLAS or cuBLAS to compute f * C * C^T.

Another obvious place for performance optimization is to not constantly up-/download data from the GPU, but instead just keep MOs, density matrix etc on the GPU to avoid data transfer latency and b/w limitations. Doing the matrix multiply efficiently is probably more important, but I don't know the timings you have.

Madu86:
@agoetz I have incorporated dgemm and cublass_dgemm calls into DIIS subroutine as we discussed. The total times now stand as follows: https://pastebin.com/mccf0uwQ. I also looked at the time taken by all DIIS components. 

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L228-L229 : 1.76 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L247-L266 : 15.59 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L272-L290 : 9.34 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L299-L324 : 15.32 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L331-L338 : 40.48 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L360-L372 : 5.69E-005 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L376-L397 : 48.31 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L383 : 11.45 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L400-L406 : 34.23 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L410-L426 : 5.39E-005 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L441-L465 : 2.19E-004 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L472-L482 : 3.84 s   

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L519-L528 : 2.82 s 

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L529-L538 : 1.77 s 

As you can see, there are several hot spots we should try to improve. But improving most of them would be tricky since we have to deal with a rank 3 tensor. Let me know what you think.

agoetz:
Thanks @Madu86 . Looking at the code blocks that you highlighted, a few things come to mind:

1. There are many copy operations that take much time and probably are not necessary. Instead, one would just swap pointers (or in old-stype programming swap integer offsets into a large array). For instance this would eliminate 40.48 s here https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L331-L338

2. Multi-dimensional array index is not optimal. For instance in arrays like `allerror(iidiis, i, j)` or `alloperator(ndiis,nbasis,nbasis)` the fastest running index is over `ndiis`. This means that the matrices are not stored contiguously in memory. Instead, data should be stored as `matrix(nbf,nbf,ndiis)`.

3. Related to point 2. use column-major array access in Fortran (row-major in C or C++). In many cases the compiler should reverse loop order, but I would not count on it, e.g. swap i and j in the example below from https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L247-L266 
```
do i = 1, nbasis
            do j = 1, nbasis
               allerror(iidiis, i, j) = quick_scratch%hold2( i, j)
            enddo
         enddo
```

4. Instead of writing loops use Fortran array intrinsics. E.g. in the example from 3. write
```
allerror(iidiis, :, :) = quick_scratch%hold2(:, :)
```
and importantly this would be better
```
allerror(:, :, iidiis) = quick_scratch%hold2(:, :)
```

Madu86:
@agoetz Thanks for the suggestions. It turned out that array access was eating up a lot of time. The times changed as follows. 

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L228-L229 : 1.94 s -> 1.76 s 

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L247-L266 : 15.59 s -> 7.06 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L272-L290 : 9.34 s -> 7.88 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L299-L324 : 15.32 s -> 5.66 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L331-L338 : 40.48 s -> 5.24 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L360-L372 : 5.69E-005 s -> 1.16E-004 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L376-L397 : 48.31 s -> 18.57 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L383 : 11.45 s -> 11.28 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L400-L406 : 34.23 s -> 1.98 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L410-L426 : 5.39E-005 s -> 1.74E-004 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L441-L465 : 2.19E-004 s -> 1.78E-004 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L472-L482 : 3.84 s -> 5.27 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L519-L528 : 2.82 s -> 2.75 s

https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L529-L538 : 1.77 s -> 1.57s

Here is the new olestra B3LYP/6-31G** final times: https://pastebin.com/uwbFxmHy

There are other places in the legacy code (e.g. scf_operator.f90) where should improve the memory access. I will make a separate PR with those changes.

agoetz (approved):
Looks all good.

agoetz:
@Madu86 Nice! Down from 570 seconds to 192 seconds with density build on GPU, and now further down to 75 seconds by improving memory access patterns. That's almost an 8x speedup! 

I am quite sure that some of the memory copy operations are not required and can be handled with pointers instead. Not sure it's worth to invest time at the moment. Have you used pointers in Fortran? It's pretty straight forward.

Can you switch loop orders over `I` and `J` here where `errormax` is computed:
https://github.com/Madu86/QUICK/blob/diisfix/src/scf.f90#L272-L290
The compiler should catch that but better safe than sorry.

Then go ahead and merge. Thanks!

Madu86:
@agoetz Thanks Andy. Ive swapped the indices you highlighted. I am merging this PR.

PR #30, opened 2020-11-12
Hot fix for atomic add

Madu86:
This pull request contains the changes required to handle large XC energies in cuda version.


agoetz (approved):
Looks good.

PR #36, opened 2021-01-14
Fixed different number of basis/primitive functions resulting from CPU and GPU OCTAGO versions

Madu86:
Turned out that I had hardcoded the cutoff in cuda code. I have fixed it in this PR.


agoetz (approved):
Looks good.

PR #37, opened 2021-01-14
Updated tests

Madu86:
I have added the test case from Issue 34 into QUICK test suite.


agoetz (approved):
Looks good. Thanks for adding the test.

Issue #38, opened 2021-01-14
Add config file for user


agoetz:
We should add a configuration file `quick.rc` in the install directory that the user can source in their shell. Content should be
```
export QUICK_INSTALL=/path/to/QUICK/installation
export QUICK_BASIS=$QUICK_INSTALL/basis
export PATH=$QUICK_INSTALL/bin:$PATH
```

The user can then do
```
source /path/to/QUICK/installation/quick.rc
```

Madu86:
I have added this capability in PR #40. Please check and merge.

PR #40, opened 2021-01-15
Added automatic quick.rc generation from configure script

Madu86:
This should fix issue #38.


agoetz (approved):
Looks good. Thanks for remembering to update LD_LIBRARY_PATH for shared library builds.

PR #41, opened 2021-01-15
Do not execute MPI_FINALIZE if running Quick through the MPI API

vwcruzeiro:


Madu86:
Everything looks good! Merging the pull request..

Issue #39, opened 2021-01-15
configure script is deleting existing libraries


vwcruzeiro:
If the Quick libraries are already installed, they are being deleted when the `configure` script is executed. This should not happen. Old libraries should be overwritten only when new ones are installed during `make install` or deleted during an action like `make uninstall`, not by the `configure` script.

Madu86:
@vwcruzeiro I have failed to reproduce this issue. In fact, I am not deleting any install/build directories from the configure script: https://github.com/merzlab/QUICK/blob/master/configure.

vwcruzeiro:
It is probably the symbolic links in $AMBERHOME/lib that are being removed then. Could you please check? Thank you!

Madu86:
I have removed symlink creation from Makefile. Instead, we are moving the libraries into amber/lib. Please check PR #42.

Madu86:
@vwcruzeiro Is this issue resolved?

vwcruzeiro:
Yes, issue is resolved in the current master branch.

PR #42, opened 2021-01-15
Eliminated symlinks in quick-amber installation

Madu86:
This should fix issue #39.


vwcruzeiro (approved):
All looks good to me.

Issue #44, opened 2021-01-19
MPI tests fail


merzlab:
MPI tests are failing on some systems when `make test` is executed. 

Testing cudampi version started on 01/18/2021 at 06:22:30 PM. 

_Running test 1 of 17
RHF energy test: s and p basis functions
grep: ene_psb5_rhf_631g.out: No such file or directory
Total energy: , Reference value: -401.844805589. Failed

Running test 2 of 17
RHF energy test: s, p and d basis functions
grep: ene_psb5_rhf_631gss.out: No such file or directory
Total energy: , Reference value: -402.013890161. Failed_

This is due to the fact that `mpirun` has been hardcoded into runtest script: https://github.com/merzlab/QUICK/blob/master/tools/runtest#L276. The choice of mpi executable and flags should be decided by the user.

Madu86:
PR #46 should fix this issue.

PR #46, opened 2021-01-20
Fix for issue 44

Madu86:
The contents in this PR fixes issue #44. Users should set mpi command using DO_PARALLEL environment variable prior to running runtest script or `make test`. For example, `export DO_PARALLEL="mpirun -np 4"` will enable runscript to use mpirun executable with 4 cpu cores. Failure to set the variable will result in serial MPI tests.


vwcruzeiro (approved):
Tested. All is good.

Issue #50, opened 2021-01-26
BLAS error in QUICK integrated AMBER


merzlab:
QUICK is looking for libblas.so when attempting to run QM/MM simulations. The paths must be correctly set or libblas.so compiled by QUICK must be moved to AMBER's lib folder.

Madu86:
PR #50 should solve this issue.

PR #52, opened 2021-01-26
Fix for issue #50

Madu86:
I have enabled moving libblas.so into AMBERHOME/lib during amber compatible quick installation. This should fix issue #50.


vwcruzeiro (changes requested):
Can we call this library libblas-quick.so after being copied to the install location?

Madu86:
done!

vwcruzeiro (approved):

Issue #67, opened 2021-01-30
QUICK/AMBER FBI based QM/MM calculations fail


merzlab:
The libblas.so file is moved into to AMBERHOME/lib as libblas-quick.so after merging PR #52. This woks fine with API based QM/MM calculations, however, calculations through file based interface are failing since quick and quick.MPI executables still look for libblas.so. This has to be fixed.

PR #68, opened 2021-01-30
Changed libblas to libblas-quick

Madu86:
This should fix issue #67.


vwcruzeiro (changes requested):
@Madu86 , don't you also need to make this name change in the place where the library is generated? It seems here that you only change the name in the place where the library is copied to the PREFIX location. Could you please double check?

Madu86:
@vwcruzeiro Oh, I made the change; but didnt commit. Should be good to go now.

vwcruzeiro:
Tested. All is good.

Issue #77, opened 2021-02-13
LAPCK/MKL diagonalization


Madu86:
For some reason, we have lost MKL and LAPACK diagonalization capability in v21.03. Internal diagonalizer is exclusively used in serial and mpi versions. Please fix this.

PR #78, opened 2021-02-14
LAPACK/MKL DSYEVD integration

Madu86:
This PR contains fixes for issue #77.

### DIFF
--- a/configure
+++ b/configure
@@ -515,6 +515,10 @@ aminstall='false'
 # control compiler output
 verbose='@'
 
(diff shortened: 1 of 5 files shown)


Madu86:
@Altaimorin Please update https://github.com/Madu86/QUICK/blob/fix77/configure#L65-L66 with prerequisites. Also update the flag list: https://github.com/Madu86/QUICK/blob/fix77/configure#L987-L991.

Issue #64, opened 2021-01-29
SCF fails with cc-pvtz basis set


merzlab:
The following calculation fails after two SCF iterations. 

Input:
HF BASIS=cc-pvtz cutoff=1.0e-9 denserms=1.0e-6  zmake ENERGY DIPOLE CHARGE=+1

C -2.74724163 -0.83655480  0.85891890
C -1.45690243 -0.47166414  0.99917288
C -0.62772841 -0.22145348 -0.15324144
C  0.68944541  0.15260156 -0.20171919
C  1.48823343  0.36448923  0.95078019
N  2.73140279  0.71794292  0.90370531
H  1.15299007  0.29662735 -1.16123028
H -1.11028708 -0.34629454 -1.10667741
H  1.08266370  0.23672479  1.93583665
H -1.04825008 -0.36804581  1.98774361
H  3.26866760  0.85959058  1.73675486
H  3.20838915  0.86445595  0.03379823
H -3.36885475 -1.02411938  1.71303219
H -3.20113376 -0.95331433 -0.10812450

Output:
 @ Begin Energy Calculation

                                         SCF ENERGY
------------------------------------------------------------------------------------------------------------------------
NCYC       ENERGY         DELTA_E      SCF_TIME  DII_CYC   DII_TIME   O_TIME  DIAG_TIME    MAX_ERR    RMS_CHG    MAX_CHG
------------------------------------------------------------------------------------------------------------------------
|  1   -250.943537035      ------         1.449     1        0.05      1.39      0.03    0.2731E+01  0.1150E+02  0.7259E+03
|  2 ****************  0.557176E+08       1.796     2        0.06      1.74      0.03    0.1515E+06  0.2701E+02  0.1714E+04
 DIIS FAILED !! PERFORM NORMAL SCF. (NOT FATAL.)
|  3 ****************  0.206646E+09       0.843     3        0.06      0.78      0.03    0.1524E+06  0.1093E+02  0.6932E+03
 DIIS FAILED !! PERFORM NORMAL SCF. (NOT FATAL.)

QUICK version: v21.03, serial executable
Compiler: GNU/9.3.1
System: Centos 7, 3.10.0-1127.18.2.el7.x86_64
CPU: Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz
Memory: 64 GB

merzlab:
The same thing happens with acetonitrile. 

Input:
[...]

PR #82, opened 2021-02-24
Improvements to Boys function

Madu86:
This PR contains updates to Boys function and should fix issue #64 in serial and MPI versions. Specifically, I have updated the cutoff for using upward recurrence relations in FmT subroutine (and FmT device kernel). The cutoff was determined by fitting Boys function values from QUICK FmT subroutine to the values from MIRP boys_exact function.
I have also added the option to link and use MIRP Boys function in the build system. MIRP library and dependencies must be compiled and MIRP_HOME, MIRP_DEP_HOME must be set before configuring QUICK. The option currently works only with --shared option.


Altaimorin:
cc-pvtz basis set needs to be added for testing.

Madu86:
Please use this: https://github.com/merzlab/QUICK/blob/QUICK-20.03/basis/CC-PVTZ.BAS. Make sure to update basis_link file accordingly.

Altaimorin:
Both serial and MPI versions have been tested and worked fine on ICER HPCC.

Madu86:
@Altaimorin Thanks! Merging the PR..

Issue #83, opened 2021-02-24
Error trap for energy calculations with f functions in CUDA version


Madu86:
As discussed in issue #64, energy calculations with f functions in CUDA version fails for some molecular systems. An error trap has to be placed to inform the user about this.

PR #84, opened 2021-02-25
Error trap for f functions

Madu86:
This PR contains an error trap to prevent users from using f functions and should fix issue #83. 

I have done this by implementing a new error handling scheme. Specifically, I have added a new module (quick_exception_module.f90, contains a couple of subroutines related to error handling) and a fortran header file (util.fh, contains some macro definitions). 
To handle new exceptions, developers must include util.fh in fortran source file and use quick_exception_module in subroutines. The error code and message must be set in the case statement in print_exception subroutine. The error code must be thrown from an appropriate place of a subroutine and caught using CHECK_ERROR() macro immediately after calling that subroutine. 
Note that exception handling will be disabled in debug/debug-time modes. This is important for stack tracing and debugging using debuggers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Brothers, E.; Götz, A.W.; Merz,K. M. QUICK-21.03 University of California San D
 
 License
 -------
-QUICK is licensed uder Mozilla Public License 2.0. More information can be found [here](https://quick-docs.readthedocs.io/en/21.3.0/license.html#mozilla-public-license-version-2-0).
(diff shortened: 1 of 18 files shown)


Altaimorin:
This fix has been tested on all four versions of QUICK and works as expected. Using CC-PVTZ basis set terminates the calculation before the initial guess and prints the error message "ERROR: f basis functions are currently not supported." in the output file. In the debug mode the error check is turned off and the run can proceed to the end.

Madu86:
@Altaimorin Thanks for testing. I am merging this PR.

Issue #87, opened 2021-02-26
Compile time error with CUDA/10.0


Madu86:
The following error is observed when QUICK is compiled with CUDA/10.0. 

```
    248 bytes stack frame, 244 bytes spill stores, 244 bytes spill loads
/mnt/i/amber-cmake/QUICK/src/libxc/maple2c_device/../../cuda/gpu_type.h(675): error: identifier "cudaErrorNotMappedAsPointer" is undefined
```

PR #88, opened 2021-02-26
Fix for issue 87

Madu86:
This PR contains a fix for the compile time error reported in issue #87. The problem was caused by usage of `cudaErrorNotMappedAsPointer ` error type which was introduced since CUDA/10.1. This has been fixed by selecting correct error type based on the CUDA toolkit version.


Madu86:
@Altaimorin Can you test the PR with several different CUDA versions on HPCC?

PR #92, opened 2021-02-27
filter out clang compiler

Altaimorin:
Fixed a typo in configure to filter out clang compiler on OS X.

### DIFF
--- a/configure
+++ b/configure
@@ -351,9 +351,10 @@ else
    echo  "Warning: Failed to detect $cc version."
 fi
 
-if [ "$os_name" = 'osx' ] && [ `echo $cc_version | grep -oE "$version_string" | wc -l` -lt 1 ] ; then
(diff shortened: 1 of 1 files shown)


Madu86:
@Altaimorin So this just filters out clang compiler. Have you tried to use clang compiler? Of course, there is no option in the configure script; but you can generate make.in file for GNU. modify it for clang and see what happens with compilation. If this works fine, we should consider adding clang option for compiler list.

Madu86:
@Altaimorin Have you checked what happens with intel compiler on OS X? @agoetz Do we know if clang, intel and gnu compilers work fine on OS X with cmake build? If yes, we should make sure configure script is also capable of doing this.

agoetz:
I have not yet tested the CMake build system on macOS. I do not have the Intel compiler install on any macOS system. I can try gnu and clang on the weekend.

agoetz:
@Madu86 Can we close this PR?

Madu86:
Yeah, this issue is no longer a concern.

Issue #93, opened 2021-02-27
QUICK fails to find the basis folder


Madu86:
QUICK_BASIS path is truncated at 80 characters and does not work for long paths. This should be fixed.

PR #99, opened 2021-02-28
Fix for long BS paths

Madu86:
I have increased the array sizes related to basis path and used fortran90 intrinsics where appropriate. This should fix issue #93.

```
| TASK STARTS ON: Sun Feb 28 10:21:46 2021
| INPUT FILE :    wat.in
| OUTPUT FILE:    wat.out
| DATE FILE  :    wat.dat
| BASIS SET PATH: /home/mmanathunga/bugfix93/quick_brown_fox_jumped_over_the_lazy_dog_again_again_again_again_again_again_again_again_again_oh_damn_again/basis

 @ Read Job And Atom

  KEYWORD=HF BASIS=6-31G CUTOFF=1.0E-9 DENSERMS=1.0E-6  ENERGY                                                                                                                                                    
 ============== JOB CARD =============
 METHOD = HATREE FOCK
 DIRECT SCF 
 SAD INITAL GUESS 
 USE DIIS SCF
 PRINT ENERGY EVERY CYCLE
 MAX SCF CYCLES =    200
 MAX DIIS CYCLES =   10
 DELTA DENSITY START CYCLE = 1000
 COMPUTATIONAL CUTOFF: 
      TWO-e INTEGRAL   =  0.100E-08
      BASIS SET PRIME  =  0.100E-19
      MATRIX ELEMENTS  =  0.100E-09
      BASIS FUNCTION   =  0.100E-09
 DENSITY MATRIX MAXIMUM RMS FOR CONVERGENCE  =  0.100E-05
 BASIS SET = 6-31G,  TYPE = CARTESIAN
| BASIS FILE = /home/mmanathunga/bugfix93/quick_brown_fox_jumped_over_the_lazy_dog_again_again_again_again_again_again_again_again_again_oh_damn_again/basis/6-31G.BAS

 @ Finish reading job

```

### DIFF
--- a/src/modules/quick_files_module.f90
+++ b/src/modules/quick_files_module.f90
@@ -240,14 +240,14 @@ module quick_files_module
         ! instant variables
         integer i,j,k1,k2
 
-        call EffChar(basisfilename,1,80,k1,k2)
-        do i=k1,k2
-            if (basisfilename(i:i).eq.'/') j=i
-        enddo
+        !call EffChar(basisfilename,1,80,k1,k2)
+        !do i=k1,k2
+        !    if (basisfilename(i:i).eq.'/') j=i
+        !enddo
 
         !write(io,'(" BASIS SET = ",a)') basisfilename(j+1:k2)
         write(io,'(" BASIS SET = ",a,",",2X,"TYPE = CARTESIAN")') trim(basisSetName)
-        write(io,'("| BASIS FILE = ",a)') basisfilename(k1:k2)
+        write(io,'("| BASIS FILE = ",a)') trim(basisfilename)
 
     end subroutine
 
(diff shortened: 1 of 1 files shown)


agoetz (approved):
Great. So much nicer to read the code now :-). I guess 320 characters should be sufficient.

Madu86:
@agoetz I just realized that `GET_ENVIRONMENT_VARIABLE ` intrinsic function that I have used here belongs to fortran standard 2003 (https://gcc.gnu.org/onlinedocs/gfortran/GET_005fENVIRONMENT_005fVARIABLE.html). Do you think we should revert this?

agoetz:
I had the same thought so had tested with a dummy program with gnu 4.8.5. It is supported. I doubt anybody will use an older compiler. I think we are good.

PR #101, opened 2021-03-01
AWG: fix runscript so MPI codes are always tested with $DO_PARALLEL

agoetz:
Some MPI and CUDA-MPI tests were executed in serial without call to mpirun. This PR fixes that.

### DIFF
--- a/tools/runtest
+++ b/tools/runtest
@@ -270,7 +270,7 @@ for buildtype in $buildtypes; do
     print_test_info "$i"
 
     # Run the test case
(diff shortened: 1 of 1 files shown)


Madu86:
@agoetz Looks good; but do we need to merge this since we have already have a second runtest script effort in PR #96?

Madu86:
@agoetz @vwcruzeiro I think this change is no longer required since we have the new test script merged into master.

agoetz:
Closing this PR as the issue was fixed in PR #108

Issue #102, opened 2021-03-01
Geometry optimization tests fail with 4 MPI tasks + 4 GPUs on Expanse


agoetz:
The tests `opt_wat_rhf_631g` and `opt_wat_rhf_ccpvdz` fail on Expanse when running with 4 MPI tasks and 4 GPUs.

- gnu 8.1.3
- openmpi 4.0.4

The output ends after `@ End Reading Molecular Information`.

The tests pass with 2  or 3 tasks / GPUs. The tests do fail if using more tasks than available GPUs.

PR #103, opened 2021-03-01
Fix for issue 102

Madu86:
GPU uploading for geometry optimization should be only performed inside optimization.f90 (https://github.com/Madu86/QUICK/blob/bugfix102/src/optimize.f90#L130-L144). But somehow this also happens also in main.f90 even if Schwarz cutoff and cutoff matrix is not computed there: https://github.com/Madu86/QUICK/blob/bugfix102/src/main.f90#L176-L202. I have eliminated the latter. This fixes the issue #102.


Madu86:
I don't see why someone would put the cutoff calculation and GPU uploading in two different places. We can simply do these tasks inside scf_operator subroutine. Anyway, lets leave it like this for now. I will make this change during the next code refactoring.

agoetz (approved):
Looks good.

agoetz:
I am going to merge this as it should be independent of any other open PRs.

vwcruzeiro:
@agoetz Did you test this PR with CUDA and CUDA+MPI?

agoetz:
I did not test it. I looked at the changes in the source code.

vwcruzeiro:
I just tested. All is good.

PR #106, opened 2021-03-02
Bugfix91

Altaimorin:
Fixed CUSolver issue for CUDA 9.0 and lower


Madu86:
Just tested this with GNU and Intel compilers and CUDA/9.0. Works fine.

Issue #115, opened 2021-03-04
Li is not identified 


Madu86:
For some reason, Lithium is not identified by QUICK. This has to be fixed. 

Input:

```
HF BASIS=6-31G** cutoff=1.0e-9 denserms=1.0e-6  zmake GRADIENT DIPOLE CHARGE=0

Li        -6.41953        1.52486        0.00000
F         -4.60901        1.52486        0.00000
```

Output:
```
 =========== Molecule Input ==========
 TOTAL MOLECULAR CHARGE  =    0    MULTIPLICITY                =    1
 TOTAL ATOM NUMBER       =    2    NUMBER OF ATOM TYPES        =    2
 NUMBER OF HYDROGEN ATOM =    0    NUMBER OF NON-HYDROGEN ATOM =    2
 NUMBER OF ALPHA ELECTRONS =    5
 NUMBER OF BETA ELECTRONS  =    4

 -- INPUT GEOMETRY -- :
    XX         -6.4195       1.5249       0.0000
    F          -4.6090       1.5249       0.0000

 -- DISTANCE MATRIX -- :
              1           2
      1     0.00000
      2     3.42139     0.00000
```

agoetz:
Looks like Li gets identified as SAD guess is done for Li, but afterwards something goes wrong as it thinks there are only 9 electrons.
```
 @ Begin SAD initial guess

 For Atom Kind =    1
 ELEMENT = LI
 BASIS FUNCTIONS =   15
 For Atom Kind =    2
 ELEMENT = F 
 BASIS FUNCTIONS =   15
```

It's even worse for LiCl, QUICK thinks there are zero electrons:
```
DFT B3LYP BASIS=def2-svp cutoff=1.0e-9 denserms=1.0e-6 GRADIENT DIPOLE

 Li         0.0 0.0 0.0
 Cl         1.5 0.0 0.0
```

```
 @ Begin SAD initial guess

 For Atom Kind =    1
 ELEMENT = LI
 BASIS FUNCTIONS =    9
 For Atom Kind =    2
 ELEMENT = CL
 BASIS FUNCTIONS =   19

 @ Finish SAD initial guess


 @ Begin Reading Molecular Information


 =========== Molecule Input ==========
 TOTAL MOLECULAR CHARGE  =    0    MULTIPLICITY                =    1
 TOTAL ATOM NUMBER       =    2    NUMBER OF ATOM TYPES        =    2
[...]

PR #118, opened 2021-03-04
Fix for issue 115

Madu86:
The issue #115 is caused by missing case conversion. In other words, If someone put atom labels as "Li" instead of "LI" QUICK is doomed. Another potential hazard is hardcoded 80s in read statements:`read (input,'(A80)')`. If someone exceeds 80 characters, we are in trouble again. We unarguably have the dumbest IO system on earth!


vwcruzeiro:
@Madu86, I see you are converting the full string to uppercase. This is good, because it makes the input case insensitive. The limit of 80 characters applies to what?

Madu86:
@vwcruzeiro Only for coordinate/point charge blocks. For job keyword, I increased this to 200 sometime ago: https://github.com/merzlab/QUICK/blob/master/src/read_job_and_atom.f90#L40.

vwcruzeiro:
Should we increase 80 to a higher number?

Madu86:
For the time being, I think we are safe. Hopefully people won't use 12 digits for each floating point number and separate columns by 12 characters.

vwcruzeiro:
Ok, ok. Let me merge this PR then.

PR #119, opened 2021-03-04
Changes in Run.tests.amber

vwcruzeiro:
Do not do extra checks for geometry optimizations, and change precision for gradient calculations.


vwcruzeiro:
I tested these changes locally, therefore I am merging.

PR #120, opened 2021-03-04
Move quick.rc to root install dir instead of bin

multiplemonomials:


agoetz (approved):
Great, works like a charm. Thanks.

Issue #85, opened 2021-02-25
Improve error handling


Madu86:
Currently, error handling in QUICK is weak. The errors must be appropriately propagated and messages must be printed in to the output.

PR #125, opened 2021-03-05
fixed the scientific notation reading issue in input

Altaimorin:
The scientific notation reading issues is fixed for all the keywords in the input file of which the value is supposed to be a double. The users are allowed to use the keywords like `cutoff=1e-9 denserms=1e-8`.

### DIFF
--- a/src/modules/quick_input_parser_module.f90
+++ b/src/modules/quick_input_parser_module.f90
@@ -42,7 +42,6 @@ module quick_input_parser_module
                 call PrtErr(OUTFILEHANDLE, "Error with keyword "//trim(keyword)//" encountered.")
                 call quick_exit(OUTFILEHANDLE,1)
             endif         
-                 
         end subroutine read_float_keyword
 
         subroutine read_integer_keyword(line, keyword, val)
--- a/src/modules/quick_input_parser_module.f90
+++ b/src/modules/quick_input_parser_module.f90
@@ -1,0 +1,83 @@
+#include "util.fh"
+module quick_input_parser_module
+
+    implicit none
+    
+    private
+    public :: read
+
+    interface read
+        module procedure read_integer_keyword
+        module procedure read_float_keyword
+        module procedure read_string_keyword
+    end interface read
+
+    contains
+
+        subroutine read_float_keyword(keywdline, keyword, val)
+            implicit none
+            character(len=*), intent(in) :: keywdline
+            character(len=*), intent(in) :: keyword
+            double precision :: val
+            integer :: i,j,ierror
+            
+            if(index(keywdline,keyword) /= 0) then
+                i = index(keywdline, trim(keyword)//trim('='))
+                if(i==0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif
+                j = scan(keywdline(i:len_trim(keywdline)), ' ', .false.)
+                read(keywdline(i+len_trim(keyword)+1:i+j-2),*, iostat=ierror) val              
+                if(ierror/=0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif                
+            endif
+        end subroutine read_float_keyword
+
+        subroutine read_integer_keyword(keywdline, keyword, val)
+            implicit none
+            character(len=*), intent(in) :: keywdline
+            character(len=*), intent(in) :: keyword
+            integer :: val
+            integer :: i,j,ierror
+
+            if(index(keywdline,keyword) /= 0) then
+                i = index(keywdline, trim(keyword)//trim('='))
+                if(i==0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif
+                j = scan(keywdline(i:len_trim(keywdline)), ' ', .false.)
+                read(keywdline(i+len_trim(keyword)+1:i+j-2),*, iostat=ierror) val
+                if(ierror/=0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif
+            endif
+        end subroutine read_integer_keyword    
+    
+        subroutine read_string_keyword(keywdline, keyword, val)
+            implicit none
+            character(len=*), intent(in) :: keywdline
+            character(len=*), intent(in) :: keyword
+            character(len=50) :: val
+            integer :: i,j,ierror
+
+            if(index(keywdline,keyword) /= 0) then
+                i = index(keywdline, trim(keyword)//trim('='))
+                if(i==0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif
+                j = scan(keywdline(i:len_trim(keywdline)), ' ', .false.)
+                read(keywdline(i+len_trim(keyword)+1:i+j-2),*, iostat=ierror) val
+                if(ierror/=0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif
+            endif
+        end subroutine read_string_keyword
+
+end module quick_input_parser_module
--- a/src/modules/quick_input_parser_module.f90
+++ b/src/modules/quick_input_parser_module.f90
@@ -16,7 +16,7 @@ module quick_input_parser_module
 
         subroutine checkformat(i,line,keyword)
             implicit none
-            integer, intent(inout) :: i
+            integer, intent(out) :: i
             character(len=*), intent(in) :: line
             character(len=*), intent(in) :: keyword
 
--- a/src/modules/quick_input_parser_module.f90
+++ b/src/modules/quick_input_parser_module.f90
@@ -20,9 +20,9 @@ module quick_input_parser_module
             character(len=*), intent(in) :: line
             character(len=*), intent(in) :: keyword
 
-            i = index(line, trim(keyword)//trim('='))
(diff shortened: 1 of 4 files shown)


src/modules/quick_method_module.f90
```diff
@@ -555,11 +567,25 @@ subroutine read_quick_method(self,keywd)
             if (index(keywd,'SCF=') /= 0) self%iscf = rdinml(keywd,'SCF')
 
             ! DM Max RMS
-            if (index(keywd,'DENSERMS=') /= 0) self%pmaxrms = rdnml(keywd,'DENSERMS')
-
-            ! 2e-cutoff
+            if (index(keywd,'DENSERMS=') /= 0) then
+                self%pmaxrms = rdnml(keywd,'DENSERMS')
```
Madu86, line 571: @Altaimorin Chi, why can't we use read function in place of rdnml? Have you checked this?

Madu86 (commented):
Is it possible to eliminate rdnml by read function in lines 541, 553 and etc.?

Altaimorin:
@Madu86 
Yes, we can eliminate all `rdnml(keywd,'$keyword')` since we already used `read`. I'm going to remove them.

Madu86:
@Altaimorin Great. One more thing, if we have a space between `=` and number (see below), we end up with a fortran error that doesn't make much sense. 

```
HF BASIS=6-31G cutoff=1.0e-9 denserms= 1e-6  zmake ENERGY DIPOLE

  O         -0.741530        1.752130        2.896280
  H         -1.111151        0.979769        3.352290
  H         -0.920500        2.036450        1.984040
```

Output: 

```
[mmanathunga@psgcluster runs]$ quick opt_wat_rhf_631g.in 
At line 564 of file quick_method_module.f90
Fortran runtime error: End of file
```
Can we handle this properly?

Altaimorin:
> @Altaimorin Great. One more thing, if we have a space between `=` and number (see below), we end up with a fortran error that doesn't make much sense.
> 
> ```
> HF BASIS=6-31G cutoff=1.0e-9 denserms= 1e-6  zmake ENERGY DIPOLE
> 
>   O         -0.741530        1.752130        2.896280
>   H         -1.111151        0.979769        3.352290
>   H         -0.920500        2.036450        1.984040
> ```
> 
> Output:
> 
> ```
> [mmanathunga@psgcluster runs]$ quick opt_wat_rhf_631g.in 
> At line 564 of file quick_method_module.f90
> Fortran runtime error: End of file
> ```
> 
> Can we handle this properly?

Good question. Let me think about that.

src/modules/quick_method_module.f90
```diff
@@ -538,13 +537,17 @@ subroutine read_quick_method(self,keywd)
             ! Density map
             ! JOHN FAVER 12/2008
             if (index(keywd,'DENSITYMAP') /= 0) then
-                self%gridspacing = rdnml(keywd,'DENSITYMAP')
+                i = index(keywd,'DENSITYMAP=',.false.)
+                j = scan(keywd(i:len_trim(keywd)), ' ', .false.)
+                read(keywd(i+10:i+j-2), *) self%gridspacing
```
agoetz, line 542: This is not robust and leads to repetitive code. You want to write a read function only once and then use it multiple times so that you will have code like

```
user input_parser, only :: read
...
if (index(input_string,'FOO') /= 0) then
   self%gridspacing = read(input_string, 'FOO')
end if
```

Then you write a module with robust functions for different data types and use interface to overload the read function.

```
module input_parser

  implicit none

  private
  public :: read

  interface read
    module procedure read_integer_keyword
    module procedure read_float_keyword
    module procedure read_string_keyword
  end interface

contains

  function read_integer_keyword(input, keyword)
    implicit none
    character(len=*), intent(in) :: input
    character(len=*), intent(in) :: keyword
    integer :: read_integer_keyword

    FOLLOWS YOUR ROBUST IMPLEMENTATION TO PARSE THE STRING

     read_integer_string = WHATEVER YOU PARSED

  function read_integer_keyword

  IMPLEMENTATION OF FUNCTIONS FOR OTHER DATA TYPES

end module input_parser
```

Of course it makes sense in the above to have a robust function internal to the parser module (i.e not public) that identifies the piece of the string that contains the data so you don't do this in every of the above functions.

Altaimorin:
@agoetz 
I'm trying to add a module `quick_input_parser_module` as you recommended:

#include "util.fh"
module quick_input_parser_module

    implicit none

    private
    public :: readkeyword

    interface readkeyword
        !module procedure read_integer_keyword
        module procedure read_float_keyword
        !module procedure read_string_keyword
    end interface readkeyword

    contains

    double precision function read_float_keyword(keywdline, keyword)
            implicit none
            character(len=200), intent(in) :: keywdline
            character(len=50), intent(in) :: keyword
            double precision :: val
            integer :: i,j,ierror

            if(index(keywdline,keyword) /= 0) then
                i = index(keywdline, trim(keyword)//trim('='))
                if(i==0) then
                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
                    call quick_exit(OUTFILEHANDLE,1)
                endif
                j = scan(keywdline(i:len_trim(keywdline)), ' ', .false.)
                read(keywdline(i+len_trim(keyword)+1:i+j-2),*, iostat=ierror) read_float_keyword
                if(ierror/=0) then
                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
                    call quick_exit(OUTFILEHANDLE,1)
                endif
            endif
        end function read_float_keyword

end module quick_input_parser_module

I added it to the Makefile and called it in `quick_method_module.f90`:
self%pmaxrms = readkeyword(keywd, 'DENSERMS')

However when compiling quick I got:
quick_method_module.f90:573:27:

             self%pmaxrms = readkeyword(keywd, 'DENSERMS')
                           1
Error: There is no specific function for the generic ‘readkeyword’ at (1)
make[2]: *** [/mnt/home/jinchi/QUICK_CMAKE/bf113_parser/build/obj/serial/quick_method_module.o] Error 1
make[1]: *** [quick_modules] Error 2
make: *** [serial] Error 2

Do you know what was going on? Thank you!

src/modules/quick_input_parser_module.f90
```diff
@@ -0,0 +1,83 @@
...
+            character(len=*), intent(in) :: keyword
+            double precision :: val
+            integer :: i,j,ierror
+            
+            if(index(keywdline,keyword) /= 0) then
+                i = index(keywdline, trim(keyword)//trim('='))
+                if(i==0) then
+                    call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
+                    call quick_exit(OUTFILEHANDLE,1)
+                endif
```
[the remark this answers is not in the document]
Altaimorin: There's two format checks per each subroutine. I was trying to replace the second check by calling the function as well but it requires negation operation on the integer `ierror`, which seems not to be supported by fortran.

src/modules/quick_input_parser_module.f90
```diff
@@ -14,70 +14,67 @@ module quick_input_parser_module
...
 
-        subroutine read_float_keyword(keywdline, keyword, val)
+        subroutine checkformat(i,line,keyword)
             implicit none
-            character(len=*), intent(in) :: keywdline
+            integer, intent(inout) :: i
+            character(len=*), intent(in) :: line
+            character(len=*), intent(in) :: keyword
+
+            i = index(line, trim(keyword)//trim('='))
```
agoetz, line 23: You don't need the second `trim` here because `'='` is guaranteed not to have any trailing blanks.

src/modules/quick_input_parser_module.f90
```diff
@@ -14,70 +14,67 @@ module quick_input_parser_module
...
+            if(i==0) then
+                call PrtErr(OUTFILEHANDLE, "USE keyword=val format in input")
```
agoetz, line 25: I am not sure what PrtErr() does. But perhaps we should indicate which keyword has an issue by adding `call PrtErr(OUTFILEHANDLE, "Error with keyword "//trim(keyword)//" encountered.")`

Madu86: Just prints the error: https://github.com/Altaimorin/QUICK/blob/bugfix113/src/subs/PrtErr.f90. Actually this is not consistent with our new error handling implementation. I will have to fix this once this PR and PR #85 is merged.

src/modules/quick_input_parser_module.f90
```diff
@@ -14,70 +14,67 @@ module quick_input_parser_module
 
     contains
 
-        subroutine read_float_keyword(keywdline, keyword, val)
+        subroutine checkformat(i,line,keyword)
             implicit none
-            character(len=*), intent(in) :: keywdline
+            integer, intent(inout) :: i
```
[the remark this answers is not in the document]
Altaimorin: `i` was declared in the functions `read_**_keyword` so shouldn't it be `inout`?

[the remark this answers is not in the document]
agoetz: Not true. Are you using the value of `i` passed into the function? No. Hence it is not `intent(in)`. All you do is set the output value, hence it is `intent(out)`. Here is the definition
https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-oneapi-dev-guide-and-reference/top/language-reference/a-to-z-reference/h-to-i/intent.html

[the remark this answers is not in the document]
Altaimorin: Oh I see. The initial value before passing is ignored. Thank you.

src/modules/quick_method_module.f90
```diff
@@ -433,8 +435,7 @@ subroutine read_quick_method(self,keywd)
             implicit none
             character(len=200) :: keyWD
             character(len=200) :: tempstring
-            integer :: itemp,rdinml,i,j
-            double precision :: rdnml
+            integer :: itemp,rdinml,i,j,ierror
```
agoetz, line 438: rdinml and ierror are not used in this subroutine.

agoetz (approved):
Looks good. @Altaimorin are you finished here or are you going to add anything else?

Altaimorin:
@agoetz 
I think for this topic nothing else will be added. Thank you!

agoetz:
I will merge this since the CI tests pass.

PR #128, opened 2021-03-07
Few more CMake fixes

multiplemonomials:
With these changes, the build of QUICK sander should finally work reliably!  It turns out that since some other changes the MPI modules weren't getting put in the right folder.


vwcruzeiro:
Tested by @multiplemonomials . Merging it.

PR #133, opened 2021-03-07
Fix for recent CI failure

Madu86:
The enclosed changes should fix the current CI failures.


Madu86:
@vwcruzeiro I have tested this. Merging now..

Issue #134, opened 2021-03-08
Bug in CUDA-MPI version


Madu86:
The following error is observed if one uses too many GPUs for molecular systems with a small number of basis functions. For instance, in this example, we have 12 basis functions and if one employs 1-2 GPUs, the calculation would run fine. The calculation would crash when we go to 3-4 GPUs simply because there is not much work to distribute among GPUs.

```
mpirun -np 4 ../../../bin/quick.cuda.MPI ene_AlH3_rhf_sto3g.in 

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x2B56D1DDB697
#1  0x2B56D1DDBCDE
#2  0x2B56FA7A63FF
#3  0x41A4F9 in mgpu_eri_greedy_distribute()
#4  0x420A19 in gpu_upload_cutoff_matrix_
#0  0x2B7C88F1E697
#1  0x2B7C88F1ECDE
#2  0x2B7CB18E93FF
#3  0x41A4F9 in mgpu_eri_greedy_distribute()
#0  0x2AD80C7B1697
#1  0x2AD80C7B1CDE
#2  0x2AD83517C3FF
#4  0x420A19 in gpu_upload_cutoff_matrix_
#3  0x41A4F9 in mgpu_eri_greedy_distribute()
#4  0x420A19 in gpu_upload_cutoff_matrix_
#0  0x2B5B964C4697
#1  0x2B5B964C4CDE
#2  0x2B5BBEE8F3FF
#3  0x41A4F9 in mgpu_eri_greedy_distribute()
#4  0x420A19 in gpu_upload_cutoff_matrix_
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun noticed that process rank 1 with PID 0 on node hsw218 exited on signal 11 (Segmentation fault).
--------------------------------------------------------------------------
```
This must be handled!

PR #135, opened 2021-03-09
Fix for issue 134

Madu86:
The enclosed changes should fix issue #134. Specifically, ERI workload won't be distributed among GPUs in multi-GPU version unless the number of basis functions is greater than a threshold.


Madu86:
I have tested this locally and I need the changes here to move forward with a second PR that I am about to place. Merging this now..

PR #136, opened 2021-03-09
Read and Write SAD guess from the input

Altaimorin:
Two flags(`readSAD` and `writeSAD`) are added which allow the users to read and write SAD guess of all types of atoms in the input.


Madu86:
@Altaimorin Nice! So I did some further improvements. Keywords `writesad` and `readsad` are helpful for us to dump sad guess for each atom of each basis set and read them for testing. For example, in the recent updates, I have added `STO-3G.SAD` folder which contains sad guess files for each atom. We need to do this for all basis sets. To create these individual atomic density file (eg. Li and F) I used following input files. 

```
HF BASIS=sto-3g cutoff=1.0e-9 denserms=1.0e-6  SCF=1 writesad energy charge=+1

  Li         -0.920500        2.036450        1.984040
```

```
HF BASIS=sto-3g cutoff=1.0e-9 denserms=1.0e-6  SCF=1 writesad energy charge=-1

  F         -0.920500        2.036450        1.984040
```

This created LI and F files in the working directory. I simply created `basis/STO-3G.SAD` and moved them there. With the other changes I made, quick should be able to read those density files from this folder. For testing, I used the following input for conventional LiF HF/STO-3G energy calculation.

```
HF  BASIS=STO-3G cutoff=1.0e-9 denserms=1.0e-6  zmake ENERGY DIPOLE CHARGE=0

Li        -6.41953        1.52486        0.00000
F         -4.60901        1.52486        0.00000
```

Then I did the same with read basis: 

```
HF  BASIS=STO-3G cutoff=1.0e-9 denserms=1.0e-6  zmake ENERGY DIPOLE CHARGE=0 readsad

Li        -6.41953        1.52486        0.00000
F         -4.60901        1.52486        0.00000
```
The readsad seems to work nicely, I got the following outputs from two inputs. 

```
                                         SCF ENERGY
------------------------------------------------------------------------------------------------------------------------
NCYC       ENERGY         DELTA_E      SCF_TIME  DII_CYC   DII_TIME   O_TIME  DIAG_TIME    MAX_ERR    RMS_CHG    MAX_CHG
------------------------------------------------------------------------------------------------------------------------
|  1   -105.074569633      ------         0.003     1        0.00      0.00      0.00    0.7391E+00  0.1585E+00  0.9877E+00
|  2   -105.227999794  0.153430E+00       0.002     2        0.00      0.00      0.00    0.1565E+00  0.1213E+00  0.4167E+00
|  3   -105.271063107  0.430633E-01       0.002     3        0.00      0.00      0.00    0.2178E+00  0.4777E-01  0.1847E+00
|  4   -105.322971633  0.519085E-01       0.002     4        0.00      0.00      0.00    0.1737E-01  0.1455E-01  0.5569E-01
|  5   -105.325324572  0.235294E-02       0.002     5        0.00      0.00      0.00    0.8921E-02  0.1343E-01  0.5236E-01
|  6   -105.326115137  0.790565E-03       0.002     6        0.00      0.00      0.00    0.3513E-02  0.1846E-02  0.6515E-02
|  7   -105.326148628  0.334907E-04       0.002     7        0.00      0.00      0.00    0.1057E-02  0.1078E-02  0.3292E-02
|  8   -105.326152539  0.391139E-05       0.002     8        0.00      0.00      0.00    0.3519E-03  0.2015E-03  0.8140E-03
|  9   -105.326152871  0.331855E-06       0.002     9        0.00      0.00      0.00    0.4543E-04  0.2395E-04  0.1230E-03
| 10   -105.326152878  0.683857E-08       0.002    10        0.00      0.00      0.00    0.1048E-04  0.1379E-04  0.7313E-04
| 11   -105.326152879  0.530827E-09       0.002     1        0.00      0.00      0.00    0.4127E-05  0.2385E-05  0.8382E-05
| 12   -105.326152879  0.522480E-10       0.002     2        0.00      0.00      0.00    0.6110E-06  0.4473E-06  0.1276E-05
------------------------------------------------------------------------------------------------------------------------
| REACH CONVERGENCE AFTER  12 CYLCES
| MAX ERROR = 0.611049E-06   RMS CHANGE = 0.447348E-06   MAX CHANGE = 0.127648E-05
 -----------------------------------------------
 ELECTRONIC ENERGY    =  -113.217689361
 CORE_CORE REPULSION  =     7.891536482
 TOTAL ENERGY         =  -105.326152879
```

```
                                         SCF ENERGY
------------------------------------------------------------------------------------------------------------------------
NCYC       ENERGY         DELTA_E      SCF_TIME  DII_CYC   DII_TIME   O_TIME  DIAG_TIME    MAX_ERR    RMS_CHG    MAX_CHG
------------------------------------------------------------------------------------------------------------------------
|  1   -105.074147273      ------         0.003     1        0.00      0.00      0.00    0.7419E+00  0.1582E+00  0.9888E+00
|  2   -105.228139549  0.153992E+00       0.002     2        0.00      0.00      0.00    0.1564E+00  0.1213E+00  0.4168E+00
|  3   -105.270930645  0.427911E-01       0.002     3        0.00      0.00      0.00    0.2181E+00  0.4783E-01  0.1849E+00
|  4   -105.322969800  0.520392E-01       0.002     4        0.00      0.00      0.00    0.1737E-01  0.1455E-01  0.5567E-01
|  5   -105.325323207  0.235341E-02       0.002     5        0.00      0.00      0.00    0.8923E-02  0.1346E-01  0.5247E-01
|  6   -105.326115096  0.791889E-03       0.002     6        0.00      0.00      0.00    0.3528E-02  0.1860E-02  0.6680E-02
|  7   -105.326148511  0.334143E-04       0.002     7        0.00      0.00      0.00    0.1110E-02  0.1080E-02  0.3317E-02
|  8   -105.326152546  0.403501E-05       0.002     8        0.00      0.00      0.00    0.3487E-03  0.1993E-03  0.8071E-03
|  9   -105.326152871  0.325233E-06       0.004     9        0.00      0.00      0.00    0.4734E-04  0.2421E-04  0.1235E-03
| 10   -105.326152878  0.711592E-08       0.004    10        0.00      0.00      0.00    0.1052E-04  0.1387E-04  0.7346E-04
| 11   -105.326152879  0.534166E-09       0.003     1        0.00      0.00      0.00    0.4185E-05  0.2411E-05  0.8443E-05
| 12   -105.326152879  0.534985E-10       0.003     2        0.00      0.00      0.00    0.6143E-06  0.4474E-06  0.1269E-05
------------------------------------------------------------------------------------------------------------------------
| REACH CONVERGENCE AFTER  12 CYLCES
| MAX ERROR = 0.614257E-06   RMS CHANGE = 0.447444E-06   MAX CHANGE = 0.126886E-05
 -----------------------------------------------
 ELECTRONIC ENERGY    =  -113.217689361
 CORE_CORE REPULSION  =     7.891536482
 TOTAL ENERGY         =  -105.326152879
```

Energies looks good. So now all you have to do is write a script the read each basis file, create QUICK input files for each atom of each basis set, run the calculations, dump the atomic densities and move them into a `basis/{BASIS}.SAD` folder.

Madu86:
Once we have sad guess for all atoms of all basis sets in place, we can consider reading atomic densities by default. With this, user won't have to worry about putting `readsad` in input.

Madu86 (changes requested):
Make sure to rebase this on recent merzlab/master.

agoetz:
Great @Madu86 and @Altaimorin . Having this in place will be good.

Altaimorin:
@Madu86 @agoetz 
Should I only considered the atom types that are covered in https://github.com/Madu86/QUICK/tree/fix109/test? I think there are only limited number (about 20?) of elements are covered in the test folder.

agoetz:
We should do this for all elements. Why should we select only a subset of supported elements? Just put a loop in the code and store the atomic densities for all atoms that are available for a given basis set. I don't know if we can drive this externally. Probably yes.

The tests don't have to cover all elements because there is nothing intrinsically different in changing the nuclear charge that could uncover a bug in the code through running the test suite. Diversity in the test suite is important to the extent that we cover different numbers of primitive and contracted functions, more compact, and more diffuse functions.

agoetz:
Now this scares me
https://github.com/merzlab/QUICK/blob/5a28f06577bcbdbcc79fe65b7a7e3f0f1d421d80/src/getMolSad.f90#L81-L93
The multiplicites are only defined a few elements. Sigh. This may be the reason why some of the SAD guesses take so long. This should be fixed first and checked whether this improves convergence (time required) for the SAD guess.

agoetz:
Also, has anybody tried a calculation with Zn? Look at this...
https://github.com/merzlab/QUICK/blob/5a28f06577bcbdbcc79fe65b7a7e3f0f1d421d80/src/getMolSad.f90#L152-L170
What the heck is that?

agoetz:
@Altaimorin @Madu86 Please first
- define correct multiplicities for all elements
- take out this special treatment of Zn
and let me know if something is not clear or if you think that I misunderstand something.

Altaimorin:
> We should do this for all elements. Why should we select only a subset of supported elements? Just put a loop in the code and store the atomic densities for all atoms that are available for a given basis set. I don't know if we can drive this externally. Probably yes.
> 
> The tests don't have to cover all elements because there is nothing intrinsically different in changing the nuclear charge that could uncover a bug in the code through running the test suite. Diversity in the test suite is important to the extent that we cover different numbers of primitive and contracted functions, more compact, and more diffuse functions.

I used a bash script to generate the input for all elements and all basis set. Then the question is what's the charge in input for all elements? Should they be all neutural? For instance:

HF BASIS=3-21G cutoff=1.0e-9 denserms=1.0e-6  SCF=1 writesad energy charge=?

AC         -0.920500        2.036450        1.984040

agoetz:
Yes, they should be neutral. But the charge in the input does not matter at all because the SAD guess is done for the neutral atoms anyways as you can see here
https://github.com/merzlab/QUICK/blob/5a28f06577bcbdbcc79fe65b7a7e3f0f1d421d80/src/getMolSad.f90#L95-L96
Please look at the source code. You have to understand what the code is doing. I am happy to answer if something is unclear but I cannot always look it up for you.

agoetz:
Wait that makes no sense what I said. That must be the nuclear charge? Please check the charge that is used for the atoms. It should be neutral.

agoetz:
But this code block answers my question what method to use for the atomic guess density:
https://github.com/merzlab/QUICK/blob/5a28f06577bcbdbcc79fe65b7a7e3f0f1d421d80/src/getMolSad.f90#L50-L54
It is set to HF. I guess we can keep that.

agoetz:
Yes, the charge gets initialized with zero. So all the new molspec objects for each atom type have a charge of zero. We are good. We always get neutral atoms.

agoetz:
What you have to do now is define the correct spin multiplicity for *all elements*. As it looks like we only get it right for alkaline metals, earth alkaline metals, column 13 (B through Tl), C, N, O, P, S, halogen atoms, and rare gas elements.

agoetz:
I see you already did take out the Zn special treatment in this PR. That's good.

After fixing the atomic spin multiplicities, I think that you can just put all supported elements into a single input file so that the SAD guess is computed and written for all elements, then stop Quick after the SAD guess. Just in case this makes it easier for you to obtain all SAD guess files.

Madu86:
@agoetz Kennie mentioned that one of his grad students studied ZN containing compounds with QUICK in past. I think this guy dumped initial guess from another code and read it here. Regarding the multiplicity, I noticed this in past but completely forgot. Great catch!

agoetz:
@Madu86 I see. Having a facility to read an initial guess is great. But what went through this grad students mind to hardcode this?

agoetz:
@Altaimorin Are you making progress with the SAD guess files? Do you need help with something? We are approaching the code freeze deadline. It is really not a good idea to make changes up to the last minute.

Madu86:
> I see. Having a facility to read an initial guess is great. But what went through this grad students mind to hardcode this?

This is the same guy who used his name for variables. Guess he had different plans for life..

Altaimorin:
@agoetz @Madu86 
Hi everyone. I just pushed new changed for SAD guess. Please take a look at `allBasAtoms` in `/basis/allBASATOMS` and run it: `./allBasAtoms`. For each basis set, a new folder for SAD guess will be created under `/basis` with the SAD output in it.

Since not all basis set have all atoms in it, you will have error messages like `** On entry to DGEMM parameter number  8 had an illegal value`. But there should be no worries.

agoetz:
@Altaimorin A few comments.
- We already have a `tools` folder. Scripts should go in there. 
- The list of elements is in the basis set files. Why run all elements and get error messages?
- If Quick really tries to compute something for an element that is not available in a basis set file, then this is a bug. The code should quit telling that there is no basis set for this element. It needs to be fixed in a separate PR. 
- Why are there new basis sets? This should be a different PR.

Altaimorin:
> @Altaimorin A few comments.

> * Why are there new basis sets? This should be a different PR.

There's no new basis sets added but just new input names for them .
I grabbed all basis set name in the basis folder and input them into the input files. For instance, `6-31G*` and `6-31G(d)` are mapped to `6-31GD.BAS` in the `basis_link` but `6-31GD` isn't. That's why I added `6-31GD` as a keyword mapping to `6-31GD.BAS` in the basis_link.

agoetz:
> For instance, 6-31G* and 6-31G(d) are mapped to 6-31GD.BAS in the basis_link but 6-31GD isn't.

The basis sets are called 6-31G* or alternatively 6-31G(d). So this is what a user should put into their input. Not 6-31GD. We should not encourage non-standard names for basis sets. But, for obvious reasons, we cannot use special characters in file names. Also, there is no need to specify upper- and lower-case versions like you introduced because we handle this internally in the Fortran parser using an to_uppercase() function (might be called differently).

basis/basis_link
```diff
@@ -13,24 +13,35 @@
...
+| #3-21G                            | 3-21G                               |
+| #6-31G                            | 6-31G                               |
+| #6-31G*                           | 6-31GD                              |
+| #6-31G(d)                         | 6-31GD                              |
+| #6-31GD                           | 6-31GD                              |
+| #6-31G**                          | 6-31GDP                             |
+| #6-31G(d,p)                       | 6-31GDP                             |
+| #6-31GDP                          | 6-31GDP                             |
+| #6-31G(2df,2pd)                   | 6-31G2DF2PD                         |
+| #6-31G2DF2PD                      | 6-31G2DF2PD                         |
```
agoetz, line 26: 6-31GDP is not a standard basis set name.

basis/basis_link
```diff
@@ -13,24 +13,35 @@
...
+| #6-311GD                          | 6-311GD                             |
+| #6-311G**                         | 6-311GDP                            |
+| #6-311G(d,p)                      | 6-311GDP                            |
+| #6-311GDP                         | 6-311GDP                            |
+| #6-311G(2df,2pd)                  | 6-311G2DF2PD                        |
+| #6-311G2DF2PD                     | 6-311G2DF2PD                        |
+| #6-311+G(2d,p)                    | 6-311PLG2DP                         |
+| #6-311PLG2DP                      | 6-311PLG2DP                         |
+| #cc-pVDZ                          | CC-PVDZ                             |
+| #CC-pVDZ                          | CC-PVDZ                             |
```
agoetz, line 39: This is a duplicate of the previous line

basis/basis_link
agoetz: Also such a change should not be in this PR as it is unrelated to enabling SAD guess read/write.

agoetz (changes requested):

Altaimorin:
@Madu86 @agoetz 
I have moved `sadguess` to `tools` folder. The `basis_link` is reverted and fixed the bug in `quick_files_module.f90` to generate the BAS and SAD paths.

tools/sadguess
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+WORKDIR=`pwd`
+BASISDIR='../basis'
```
agoetz, line 3: Absolute paths are dangerous. We set QUICK_INSTALL and QUICK_BASIS in quick.rc. Better check if these are defined and use those.

tools/sadguess
```diff
@@ -0,0 +1,80 @@
...
+	basname=${basmap[$BAS]}
+	dirname="$BASISDIR/$BAS.sadinput"
+
+	if [  -d "$dirname" ];then
+		rm -r "$dirname"
+	fi
+	mkdir -p $dirname
+
+	cd $dirname
+	cp ../../bin/quick .
```
agoetz, line 45: Quick should be in the path, no need to copy here.

agoetz:
@Altaimorin I tried the script. It works fine but I don't think it is a good idea to automatically run for all basis sets and to modify the sad files in the basis directory. I would make the script use the basis set file name as command line parameter and stay in the local directory where the user is executing the script. Imagine I want to add a new basis set, let's say I have added $QUICK_INSTALL/basis/6-31GD.bas. This is what I would want to do:
```
mkdir foo; cd foo
sadguess 6-31GD.bas
# Now I have a bunch of files X.in, X.out, X for each element X
# I check if everything looks good and if yes I will do
rm *.in *.out
mkdir $QUICK_BASIS/6-31GD.sadinput
mv * $QUICK_BASIS/6-31GD.sadinput
git add $QUICK_BASIS/6-31GD.sadinput
git commit
```

If you change this, make sure the script gets installed in $QUICK_INSTALL/bin, and fix the spin multiplicities as described in one of my comments above, then we can merge this PR.

In a follow-up PR we can then add the SAD guess files and enable reading of SAD guess by default.

Altaimorin:
> @Altaimorin I tried the script. It works fine but I don't think it is a good idea to automatically run for all basis sets and to modify the sad files in the basis directory. I would make the script use the basis set file name as command line parameter and stay in the local directory where the user is executing the script. Imagine I want to add a new basis set, let's say I have added $QUICK_INSTALL/basis/6-31GD.bas. This is what I would want to do:
> 
> ```
> mkdir foo; cd foo
> sadguess 6-31GD.bas
> # Now I have a bunch of files X.in, X.out, X for each element X
> # I check if everything looks good and if yes I will do
> rm *.in *.out
> mkdir $QUICK_BASIS/6-31GD.SAD
> mv * $QUICK_BASIS/6-31GD.SAD
> git add $QUICK_BASIS/6-31GD.SAD
> git commit
> ```
> 
> If you change this, make sure the script gets installed in $QUICK_INSTALL/bin, and fix the spin multiplicities as described in one of my comments above, then we can merge this PR.
> 
> In a follow-up PR we can then add the SAD guess files and enable reading of SAD guess by default.

Sorry I'm a bit confused about the spin multiplicities stuff. It's said `For example, the ground state of the carbon atom is a 3P state. ` but in QUICK the multiplicity of phosphorus was set to 4. Is there a source to look for it?

agoetz:
>Sorry I'm a bit confused about the spin multiplicities stuff. It's said For example, the ground state of the carbon atom is a 3P state. but in QUICK the multiplicity of phosphorus was set to 4. 

Yes, the ground state of carbon is 1s2 2s2 2p2 with two electrons with same spin in two different p orbitals, all other electrons paired, resulting in a 3P triplet state. Higher energy states would be singlet states with those two electrons in two different p orbitals or the same p orbital giving 1D and 1S states (I think). So what Quick has currently is correct
https://github.com/merzlab/QUICK/blob/e7afea26fd7bf170c5c9b0a51dac19212637c08b/src/getMolSad.f90#L91

What does this have to do with phosphorus?

>Is there a source to look for it?

You can probably figure out all configurations by looking at the periodic table. It's been a while for me but you probably can find the atomic ground state electron configurations in any chemistry textbook.

agoetz:
Those spin multiplicities for atomic electronic ground states should go into `quick_constants_module.f90`.

Makefile
```diff
@@ -68,6 +68,7 @@ checkfolders:
 .PHONY: noinstall install serialinstall mpiinstall cudainstall cudampiinstall aminstall
 
 install: $(INSTALLTYPES)
+	@cp -f $(toolsfolder)/sadguess $(exefolder)
```
agoetz, line 71: This should probably be in the serial install target since you use the serial quick executable for the guess generation.

src/modules/quick_constants_module.f90
```diff
@@ -99,6 +99,19 @@ module quick_constants_module
    & 'TL','PB','BI','PO','AT','RN','FR','RA','AC','TH', &
    & 'PA','U '/)
 
+    integer, dimension(0:92) :: SPINMULT = &
+   & (/0, 2, 1, 2, 1, 2, 3, 4, 3, 2, 1, &
```
agoetz, line 103: Looks good. Thanks! But do we need a zero element? I would arrange as follows as it is easy to check:
```
 (/0, &
   2,                                                 1, &
   2, 1,                               2, 3, 4, 3, 2, 1, &
   2, 1,                               2, 3, 4, 3, 2, 1, &
   2, 1, 2, 3, 4, 7, 6, 5, 4, 3, 2, 1, 2, 3, 4, 3, 2, 1, &
   2, 1, 2, 3, 6, 7, 6, 5, 4, 1, 2, 1, 2, 3, 4, 3, 2, 1, &
   2, 1, &
     2, 1, 4, 5, 6, 7, 8, 9, 6, 5, 4, 3, 2, 1, &
         2, 3, 4, 5, 6, 5, 4, 3, 2, 1, 2, 3, 4, 3, 2, 1, &
   2, 1, 2, 3, 4, 5/)
```
I think you do not need the ampersand at the beginning of the continuation line. See e.g. here https://pages.mtu.edu/~shene/COURSES/cs201/NOTES/chap01/continue.html

tools/sadguess
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
-WORKDIR=`pwd`
-BASISDIR='../basis'
+WORKDIR=$QUICK_INSTALL/tools
```
agoetz, line 2: This line seems superfluous. We also don't have a tools directory in the installation directory afaik.

tools/sadguess
```diff
@@ -34,19 +36,18 @@ done
...
 	basname=${basmap[$BAS]}
-	dirname="$BASISDIR/$BAS.sadinput"
+	dirname="$CURDIR/$BAS.sadinput"
 
 	if [  -d "$dirname" ];then
 		rm -r "$dirname"
 	fi
 	mkdir -p $dirname
 
 	cd $dirname
```
agoetz, line 46: Can this work? You go into a directory within the loop but you never leave the directory. You need a `cd ..` somewhere. But this is quite error prone. A safer way to do this is to use pushd and popd, e.g.
```
pushd $dirname
... do stuff ...
popd
```

agoetz (changes requested):

tools/sadguess
Altaimorin: This worked since at line 73 there's `cd $WORKDIR`. That's why I put `WORKDIR=$QUICK_INSTALL/tools` at the top. It has been switched `pushd` and `popd` in the new commit.

agoetz (approved):
@Madu86 If the CI tests pass then we can merge this. A few points should or could be improved
- error handling: throw error if both RDSAD and WRSAD is used.
- error handling: throw error if RDSAD but SAD file not found.
- install sadguess script with CMake
Also keep in mind that once we add all SAD guess files and activate RDSAD by default, then we should add logic to deactivate the SAD guess. Maybe a new keyword COMPUTESAD.

agoetz:
I have made these modifications myself and pushed to this PR because we are running out of time.

agoetz (changes requested):
There is still a bug. It's in the call of getenergy(). I am fixing this.

agoetz (approved):
This now looks all good to me. The tests pass locally on the machine I worked on. Since all CI tests pass I will merge this.

PR #138, opened 2021-03-09
Error handling improvements

Madu86:
This PR contains changes necessary to propagate and handle exceptions properly in QUICK. Specifically, I have introduced `ierr` variable to subroutine calls (eg. https://github.com/Madu86/QUICK/blob/fix85/src/main.f90#L63), throw exceptions from appropriate places by assigning integer values to ierr variable (eg. https://github.com/Madu86/QUICK/blob/fix85/src/subs/quick_open.f90#L84. The exception are caught by CHECK_ERROR() or SAFE_CALL() macros (eg. https://github.com/Madu86/QUICK/blob/fix85/src/main.f90#L64, https://github.com/Madu86/QUICK/blob/fix85/src/main.f90#L60) which are defined in a header file: https://github.com/Madu86/QUICK/blob/fix85/src/util/util.fh#L35-L48. Exception handling is done inside a new module: https://github.com/Madu86/QUICK/blob/fix85/src/modules/quick_exception_module.f90. @vwcruzeiro Note that I have also added `ierr` variables to API and you will have to update the function calls from Sander (see here for eg: https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L68, https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L104, https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L116, https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L136). These changes should help us to handle most important exceptions in QUICK at the moment; but of course we will have to further improve this implementation. I think this is a good foundation. Let me know your suggestions.


src/util/util.fh
```diff
@@ -14,9 +14,20 @@
...
+#define CPHFFILEHANDLE   2024
+#define BASISFILEHANDLE  2025
+#define ECPFILEHANDLE    2026
+#define BASISCFILEHANDLE 2027
+#define PDBFILEHANDLE    2028
+#define DATAFILEHANDLE   2029
+#define INTFILEHANDLE    2030
 
 ! For the following definitions to work, one must use quick_exception_module
 ! in a subroutine. For MPI/CUDAMPI versions, quick_mpi_module must be used.
```
agoetz, line 33: In the line below, do you know why we don't do checks in the DEBUG version, i.e. when DEBUG is defined? That seems strange.

agoetz: I wasn't able to highlight the line as this is not a line in the code that was changed.

agoetz (commented):
Looks all good to me. This is a good start. I compiled on chinotto with gnu 4.8.5 and openmpi 4.0.5 and cuda 11.0 using CMake. Compilation works.

@vwcruzeiro  will have to update the API calls. Good to merge once you have done this.

I started the tests and quick.cuda.MPI works so far but I have not waited for tests to end.

This is related to a different PR, but tests take too long. We need a flag for short test suite to run only some of the tests. Also I get weird binary stuff at the beginning of my test log file:
```
[agoetz@chinotto] source install/quick.rc
[agoetz@chinotto] cd install
[agoetz@chinotto] nohup ./runtest > runtest.log
[agoetz@chinotto] less runtest.log
"runtest.log" may be a binary file.  See it anyway? 
```
Here is the beginning of `runtest.log` as seen with `less`:
```
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@Testing QUICK-21.03 cudampi version started on 03/08/2021 at 11:17:24 PM. 

Running test 1 of 145
```

Madu86:
> > This is related to a different PR, but tests take too long. We need a flag for short test suite to run only some of the tests. Also I get weird binary stuff at the beginning of my test log file:
> 
> [agoetz@chinotto] source install/quick.rc
> [agoetz@chinotto] cd install
> [agoetz@chinotto] nohup ./runtest > runtest.log
> [agoetz@chinotto] less runtest.log
> "runtest.log" may be a binary file.  See it anyway? 
> Here is the beginning of runtest.log as seen with less:
> 
> ^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@Testing QUICK-21.03 cudampi version started on 03/08/2021 at 11:17:24 PM. 
> 
> Running test 1 of 145

Oh, never had those smiley faces in my runtest.log files. Let me check what's going on.

agoetz:
> Can you please confirm if you have seen these strange characters once the runtest.log is fully written? I couldn't reproduce this.
Not reproducible. Must have been something I did. Sorry for the red herring.

Madu86:
@agoetz Ok. Thanks for confirming. @vwcruzeiro Can you please make the necessary changes in Amber interface, hook up this version and test QM/MM?

vwcruzeiro:
@Madu86 Yes, I can. I will do that after this PR is merged. Please show me examples of how the API functions should be called now and how `ierr` should be used. Thanks.

Madu86:
@vwcruzeiro  You just have to pass an integer variable (initialized to zero) when you call setQuickMPI, setQuickJob, getQuickEnergyGradients and deleteQuickJob. See here: https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L68, https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L104, https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L116, https://github.com/Madu86/QUICK/blob/fix85/src/quick_api_test.f90#L136 for an example.

Madu86:
@agoetz If you don't have other concerns, please merge this PR so that @vwcruzeiro can proceed.

agoetz:
@Madu86  Why do we pass a status variable `ierr` in the API if we do not use it? Shouldn't we catch QUICK library call errors in sander by using the error module and calling CHECK_ERROR() or something along those lines?

Madu86:
@agoetz I thought Amber has its own ways of handling errors when err /= 0.

agoetz:
We have two ways of handling errors in sander. Either call sander_bomb() with a message like "QUICK had an error", but that really just stops the program. Or use an assertion macro `ASSERT(ierr == 0)`, which will stop the program with an error message what type of assertion failed in which file at which line.

Either case is not very informative. Instead, it would be helpful if the user knew what actually went wrong.

Madu86:
I see. I cannot update API example with CHECK_ERROR() macro since this requires me to include `quick_exception_module` and `quick_mpi_module`. The idea of this file is to show users how to call QUICK and adding these error handling stuff would make things complicated. Instead, I would put something like `if(err /=0) exit(1)` and keep things simple in this example. @vwcruzeiro can handle errors consistently with Amber.

agoetz:
OK. So you suggest that Vinicius uses the CHECK_ERROR() macro in the module that he wrote for sander?

Madu86:
@agoetz Yes lets do that.

vwcruzeiro:
@Madu86 , I think you initial suggestion for `ierr` is sufficient. We will be able to tell the users which exact function failed (setQuickMPI, setQuickJob, getQuickEnergyGradients or deleteQuickJob) in the Amber output file. This should be enough. Users can consult quick.out for more specific information.

Madu86:
Commit # 1d96abad83ede3873fca348d894d2506348a6423 has introduced a bug into API. Let me fix that.

vinicius:
Please, somebody remove me from this thread. You've mistaken me for someone
else (maybe Vinicius Cruzeiro)

Em qua., 10 de mar. de 2021 às 03:52, Andreas Goetz <
***@***.***> escreveu:

Madu86:
@vinicius Our apologies.. I have removed you from this thread.

Issue #142, opened 2021-03-10
QUICK doesn't read user specified SCF cycles anymore 


Madu86:
@agoetz @Altaimorin The following doesn't work anymore. 

 @ Read Job And Atom

```
  KEYWORD= DFT LIBXC=GGA_X_B88,GGA_C_LYP BASIS=6-31G DIPOLE CUTOFF=1.0D-10 DENSERMS=1.0D-6 CHARGE=+1 SCF=1                                                                          

| Error: Error with keyword SCF encountered.

| Error Termination. Task Failed on: Wed Mar 10 09:06:17 2021
```

This must be fixed ASAP.

agoetz:
@Altaimorin you are searching for a space to terminate the value field. This obviously only works when the value to be read is not at the end of the line. My bad, I did not notice that when reading the code and unfortunately the CI tests did not catch this. @Altaimorin Please make the parser robust and try to break it with many, many tests.

Issue #140, opened 2021-03-10
Geometry optimization fails 


Madu86:
After the most recent merge, the OPTIMIZE keyword is not properly read. The following fails: 

```
 @ Read Job And Atom

  KEYWORD=DFT LIBXC=HYB_GGA_XC_PBEH BASIS=DEF2-SVP CUTOFF=1.0E-9 DENSERMS=1.0E-6 OPTIMIZE DIPOLE CHARGE=+1                                                                                             

| ERROR: Error with keyword OPTIMIZE encountered.

| Error Termination. Task Failed on: Tue Mar  9 21:51:27 2021
```

But this one doesn't fail. 

```
@ Read Job And Atom

  KEYWORD=HF BASIS=6-31G CUTOFF=1.0E-9 DENSERMS=1.0E-6  ZMAKE OPTIMIZE=5 DIPOLE                                                                                                                        
 ============== JOB CARD =============
 METHOD = HATREE FOCK
 DIRECT SCF
 SAD INITAL GUESS
 USE DIIS SCF
 PRINT ENERGY EVERY CYCLE
 Z-MATRIX CONSTRUCTION
 DIPOLE
 GEOMETRY OPTIMIZATION NUMERICAL GRADIENT
 REQUEST CRITERIA FOR GEOMETRY CONVERGENCE:
      MAX ALLOWED GEO CHANGE  =  0.189E+00
      MAX GEOMETRY CHANGE     =  0.180E-02
      GEOMETRY CHANGE RMS     =  0.120E-02
      MAX GRADIENT CHANGE     =  0.100E-02
      GRADIENT NORMALIZATION  =  0.300E-03
 GRADIENT CALCULATION
 MAX SCF CYCLES =    200
 MAX DIIS CYCLES =   10
 DELTA DENSITY START CYCLE = 1000
 COMPUTATIONAL CUTOFF:
      TWO-e INTEGRAL   =  0.100E-08
      BASIS SET PRIME  =  0.100E-19
      MATRIX ELEMENTS  =  0.100E-09
      BASIS FUNCTION   =  0.100E-09
      GRADIENT CUTOFF  =  0.100E-08
 DENSITY MATRIX MAXIMUM RMS FOR CONVERGENCE  =  0.100E-05
 BASIS SET = 6-31G,  TYPE = CARTESIAN
| BASIS FILE = /home/mmanathunga/fix109/basis/6-31G.BAS

 @ Finish reading job


 @ Begin SAD initial guess

[...]

PR #143, opened 2021-03-11
new input parser

Altaimorin:
A new input parser is implemented. It allows the users to have spaces on the two sides of the equal sign which maps the keyword to the value like `OPTIMIZE = 5` as in https://github.com/merzlab/QUICK/issues/140. It can also parse the keyword at the end of the input line with no trailing spaces (as required in https://github.com/merzlab/QUICK/issues/142).

### DIFF
--- a/src/modules/quick_input_parser_module.f90
+++ b/src/modules/quick_input_parser_module.f90
@@ -14,18 +14,48 @@ module quick_input_parser_module
 
     contains
 
-        subroutine checkformat(i,line,keyword)
+        subroutine trimSpace(i,j,line,keyword,found)
             implicit none
-            integer, intent(out) :: i
+            integer, intent(out) :: i,j
             character(len=*), intent(in) :: line
             character(len=*), intent(in) :: keyword
+            logical, intent(out) :: found
 
-            i = index(line, trim(keyword)//'=')
-            if(i==0) then
-                call PrtErr(OUTFILEHANDLE, "Error with keyword "//trim(keyword)//" encountered.")
-                call quick_exit(OUTFILEHANDLE,1)
+            print *, "keyword is ", keyword
+    
+            !first, go to the right to the end of the keyword
+            i = index(line, trim(keyword))+len_trim(keyword)
+            print *, "line(i:i+4) is", line(i:i+4)
+        
+            !ignore all spaces left to the equal sign
+            do while(line(i:i)==' ')
+                i=i+1
+            end do
+
+            !if equal sign not found, return
+            if(line(i:i) /= '=') then
+                print *, "didn't find assigned value"
+                found = .false.
+                return
             endif
-        end subroutine checkformat
+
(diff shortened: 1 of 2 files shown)


src/modules/quick_input_parser_module.f90
```diff
@@ -14,18 +14,48 @@ module quick_input_parser_module
...
-                call PrtErr(OUTFILEHANDLE, "Error with keyword "//trim(keyword)//" encountered.")
-                call quick_exit(OUTFILEHANDLE,1)
+            print *, "keyword is ", keyword
+    
+            !first, go to the right to the end of the keyword
+            i = index(line, trim(keyword))+len_trim(keyword)
+            print *, "line(i:i+4) is", line(i:i+4)
+        
+            !ignore all spaces left to the equal sign
+            do while(line(i:i)==' ')
```
agoetz, line 31: Would be safer to also ignore tab characters. You never know what users will do.

agoetz (changes requested):
OK. So this looks good - it looks like it will allow users to do any of
```
foo=1
foo =1
foo= 1
foo = 1
```
including having a keyword value pair as last element on the line.

Changes required
1) remove debug print statements

2) We need to distinguish keys with required and optional values. E.g. `basis` or `cutoff` require a value and not putting a value would constitute an input error. The program must quit with an error message. This is different for the `optimize` keyword, which has an optional value. The current implementation makes all values optional.

agoetz:
I think all you have to do is to pass down an optional logical argument, e.g. `required`, that you set to `.true.` by default. And then when you call the parser function for `optimize`, you do something like `call read(line, `optimize`, iopt, required=.false.)`. Now, if you do that, you will have to give `intent(inout)` to the dummy parameter where you return the value (if found). You should not use `intent(out)` because according to the Fortran standard the variable is undefined upon entry (you won't notice this for an integer or floating point number, but would for a pointer or derived data type).

src/modules/quick_input_parser_module.f90
```diff
@@ -34,29 +64,34 @@ subroutine read_float_keyword(line, keyword, val)
             character(len=*), intent(in) :: keyword
             double precision :: val
```
agoetz, line 65: You should add `intent(out)` here. If assignment is optional, and `val` could have a previously defined value (e.g. max number of optimization cycles set before), then you should use `intent(inout)` here.

agoetz (approved):
This looks like it should work. Have you tested this? If it compiles and works as expected, then I think we can merge.

src/modules/quick_input_parser_module.f90
Altaimorin: > This looks like it should work. Have you tested this? If it compiles and works as expected, then I think we can merge.

Yes I tested on my side with both correct and erroneous inputs. Worked as expected.

agoetz:
OK. I will merge this.

PR #149, opened 2021-03-12
Update CMake build system compiler flags

agoetz:
Update defaults
- disable warnings so users don't get shocked
- warnings can be enabled with -DWARNINGS=TRUE
- downgrade frome -O3 to -O2 so we won't have nasty surprises with the release on untested platforms
- remove duplicate -O2 flag
- update version number to 21.03
- remove PGI, CRAY, CLANG as these have not been tested or are not working

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ if(NOT INSIDE_AMBER)
 	set(BUNDLED_3RDPARTY_TOOLS blas lapack)
 	include(3rdPartyTools)
 	
-	include(CompilerFlags)
+	include(QUICKCompilerFlags)
 
 	#CPack setup
 	# --------------------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.9.0) # version 3.9.0 needed for FindMPI and FindOpenMP versions which provide imported targets
 
-project(quick LANGUAGES NONE VERSION 20.03)
+project(quick LANGUAGES NONE VERSION 21.03)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/quick-cmake)
 if(NOT INSIDE_AMBER)
@@ -44,15 +44,7 @@ option(NOF "Disables the compilation of QUICK's time consuming f functions in th
 set_property(DIRECTORY . PROPERTY COMPILE_DEFINITIONS $<$<CONFIG:Debug>:DEBUG>)
 
 
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
-
-	# QUICK needs to be built with -O2 not Amber's default of -O3
-	if(OPTIMIZE)
-		list(APPEND OPT_CFLAGS -O2)
-		set(OPT_CFLAGS_SPC "${OPT_CFLAGS_SPC} -O2")
-	endif()
-
-elseif("${CMAKE_C_COMPILER_ID}" STREQUAL "Intel")
+if("${CMAKE_C_COMPILER_ID}" STREQUAL "Intel")
 	set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -traceback")
 endif()
 
@@ -61,20 +53,17 @@ if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "GNU")
 	set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -cpp")
 	add_definitions(-DGNU)
 
-	if(OPTIMIZE)
-		list(APPEND OPT_FFLAGS -O2)
-		set(OPT_FFLAGS_SPC "${OPT_FFLAGS_SPC} -O2")
-	endif()
-
 	if(CMAKE_Fortran_COMPILER_VERSION VERSION_GREATER_EQUAL 10)
 		set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fallow-argument-mismatch")
 	endif()
 
 endif()
+
 if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "Intel")
 	set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -cpp -diag-disable 8291")
 	set(CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} -traceback")
 endif()
+
 if("${CMAKE_Fortran_COMPILER_ID}" STREQUAL "PGI")
 	set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -cpp") # force use of preprocessor for Fortran files with the "wrong" file extension
 endif()
--- a/cmake/AmberCompilerConfig.cmake
+++ b/cmake/AmberCompilerConfig.cmake
@@ -10,11 +10,8 @@ set(COMPILER_HELP "
                  -----------------------------------------------------------------------
   COMPILER value | C executable | C++ executable | Fortran executable | tested versions
   --------------------------------------------------------------------------------------
-      GNU        |     gcc      |      g++       |     gfortran       | 4.4.7, 4.8.4 +
-      INTEL      |     icc      |      icpc      |     ifort          | 12 - 17
-      PGI        |     pgcc     |      pgc++     |     pgf90          | 14.9, 15.4, 16.5
-      CLANG      |     clang    |      clang++   |     gfortran       | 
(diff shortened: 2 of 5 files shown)


agoetz:
Hi @Madu86 Can you please add @multiplemonomials as reviewer? You have to add invite him as collaborator first under settings for the repo I think.

agoetz:
@multiplemonomials I made a few minor changes to the CMake build. It works as I intended and tests pass for all versions, serial, MPI, cuda, cuda.MPI, but can you please take a quick look to make sure I did not do something that should not be done. Thanks very much!

multiplemonomials:
Hmm, these changes look good on their own.  The issue is, any changes you make in the cmake/ folder will not get used when QUICK is built by Amber (because the top-level CMake scripts in Amber are used instead).  That's why I added the optimization flags into the top-level CMakeLists.txt -- are those flags not correct?

Also, have you tried passing `-DUNINITIALIZED_WARNINGS=FALSE` to CMake?  It should get rid of a lot of the common ones.

agoetz:
> The issue is, any changes you make in the cmake/ folder will not get used when QUICK is built by Amber (because the top-level CMake scripts in Amber are used instead)

Hm. We probably should change that. Perhaps after the release since there is not much time. We should have different rules for different packages, e.g. Quick, cpptraj etc. And may want to look into different optimization flags for specific files.

> That's why I added the optimization flags into the top-level CMakeLists.txt -- are those flags not correct?

I was wondering about that. It works technically to override the first optimization flag but it looks weird to have `-O3 -mtune=native -O2`. We also had `-O2` only for GNU C and Fortran, not CXX and not Intel.

> Also, have you tried passing -DUNINITIALIZED_WARNINGS=FALSE to CMake? It should get rid of a lot of the common ones.

Yes - this still leads to lots of warnings. We will need to fix those but this won't happen before the Amber release. And many warnings are out of our control because they come from libxc. The best is to disable all warnings for Quick. Developers can switch them on if they want.

Also, we don't want to allow builds of Quick with PGI, CLANG, CRAY (not working or not tested). We should throw a warning if a user tries to configure for those compilers.

Madu86:
> Can you please add @multiplemonomials as reviewer? You have to add invite him as collaborator first under settings for the repo I think.

Oh, I thought he is already in the collaborator list. Let me fix that.

Madu86:
@agoetz Done. @multiplemonomials should receive an invitation soon.

multiplemonomials:
> Hm. We probably should change that. Perhaps after the release since there is not much time. We should have different rules for different packages, e.g. Quick, cpptraj etc. And may want to look into different optimization flags for specific files.

CMake does allow different rules for different packages, you just have to specify them in the root CMakeLists.txt and not in the cmake/ folder.  Sounds like there were some issues with that, so I can try to adjust that code if needed.

agoetz:
> CMake does allow different rules for different packages, you just have to specify them in the root CMakeLists.txt and not in the cmake/ folder. Sounds like there were some issues with that, so I can try to adjust that code if needed.

Sounds like we need the complete compiler flag definitions in the root CMakeLists.txt then. We don't want to use any global Amber default. This also means that the file cmake/CompilerFlags.cmake should be removed (and probably others when we find time?) although it is nice to have this in a separate file that is included in the root CMakeLists.txt. I thought there is a way to package a project (Quick) for integration with other CMake projects (Amber).

multiplemonomials:
> We don't want to use any global Amber default. 

Wait why is this?  I assume we'd want to use the standard Amber compiler flags because Amber already did the hard part of finding out what flags are the best for HPC on each compiler and platform.

agoetz:
> Wait why is this? I assume we'd want to use the standard Amber compiler flags because Amber already did the hard part of finding out what flags are the best for HPC on each compiler and platform.

I do not think so. Those are flags that generally work for pmemd and other codes in Amber. But afaik those are not global. In the legacy build system we have e.g. PMEMD specific build flags. I do not know what cpptraj does for example. QUICK is a completely different beast. It must build on its own and is unlikely to be happy with flags that may work for sander or pmemd.

multiplemonomials:
Hmm.  Well pmemd actually has a whole set of dedicated flags that are applied only for it, I copied the legacy build system exactly on that one.  The flags I'm talking about are the general flags that end up in Amber's CFLAGS and FFLAGS variables.

agoetz:
> pmemd actually has a whole set of dedicated flags that are applied only for it, I copied the legacy build system exactly on that one

How is this handled? I have not seen those reported. Developers who work with pmemd probably will want to see those and have the ability to change those.

Is there no easy way to handle Quick as a package? This is what I think to remember that you mentioned but I may very well have misunderstood what this means. My CMake expertise is rather limited.

multiplemonomials:
> Is there no easy way to handle Quick as a package? 

What exactly do you mean by this?  You mean like have it not inherit any compile flags from Amber?  Yes this is possible using ExternalProject, but it's a huge amount of work to set up and verify, especially since libraries from QUICK need to be used by Amber.

multiplemonomials:
> How is this handled?

This is done in amber/cmake/PMEMDCompilerFlags.cmake

agoetz:
>  You mean like have it not inherit any compile flags from Amber? Yes this is possible using ExternalProject, but it's a huge amount of work to set up and verify, especially since libraries from QUICK need to be used by Amber.

Yes. Quick is a completely independent package. We also want to ship it with Amber so we can get it into the hands of more users. A developer should be able to compile their own version of Quick (not inside of Amber) and link sander against the dev version.

I do not know how ExternalProject works. Why is this a huge amount of work? All that sander needs for linking is the include path, library path and library name.

agoetz:
> This is done in amber/cmake/PMEMDCompilerFlags.cmake

Then we could use QUICK/cmake/QUICKCompilerFlags.cmake, right? And later we can remove everything that is Amber specific as it should not be in QUICK.

agoetz:
I added a cmake/QUICKCompilerFlags.cmake and include this in place of CompilerFlags in the root CMakeLists.txt. It works as intended. It may be more convoluted than need be. I do not understand the difference between OPT_CFLAGS, CMAKE_C_FLAGS, CMAKE_C_FLAGS_RELEASE etc. and when or how e.g. OPT_FFLAGS gets mapped into CMAKE_Fortran_FLAGS. Perhaps we can simplify all of this down the road.

I started with cmake/CompilerFlags.cmake and changed / removed stuff that is obsolete (checking for 386 architecture, GNU compiler versions < 4.2, dragonegg (?)) or not tested for QUICK (Windows, PGI, CRAY, CLANG). I left in many other things of which I do not know what they do or if they are required. They may not be required because I do not see equivalents in PMEMDCompilerFlags.cmake.

There are still some modifications to compiler definitions in the root CMakeLists.txt that I did not move into QUICKCompilerFlags.cmake, mostly because I do not know how some of this works.

As a note aside, In Amber, we need to quit if someone chooses PGI, CRAY or CLANG and enables QUICK.

agoetz:
Can we use this root CMakeLists.txt and cmake/QUICKCompilerFlags.cmake for the upcoming release?

We can then simplify the Quick CMake build system after the release
- remove anything that is not required by Quick, i.e. keep it as lean as possible
- include it as ExternalProject into Amber

multiplemonomials:
Dr. Goetz, we only have 3 days left before the Amber code freeze.  I am not making any changes that are not essential, and you said yourself that QUICK is passing all tests even with the current configuration.  If you want to merge it and test it, go ahead, but the only thing I'm still signed up to do is fix that sander segfault.

Setting up an ExternalProject integration takes, on average, a full day or two of work.  You have to set up command line options to proxy all of the libraries (BLAS and LAPACK) from Amber's build system into QUICK.  Then you have to have QUICK build.  Then you have to set up CMake targets that will represent, in advance, all the libraries that quick _will generate_ (including accounting for their different naming on multiple platforms).  Finally, you have to import those libraries back into the upper-level CMake build system, including setting up their module dirs, link dependencies, etc.  It's not for the faint of heart, and I do not have this kind of time anytime soon.

agoetz:
>Dr. Goetz, we only have 3 days left before the Amber code freeze. I am not making any changes that are not essential, and you said yourself that QUICK is passing all tests even with the current configuration. If you want to merge it and test it, go ahead, but the only thing I'm still signed up to do is fix that sander segfault.

In this case I will attempt to put all compiler flags into the root CMakeLists.txt so Quick has control over its build flags. We must replicate the flags from Quick's configure script. Only those have been well tested over a longer period of time. More aggressive optimization may well lead to issues on hardware/compiler versions that we have not tested on.

> Setting up an ExternalProject integration takes, on average, a full day or two of work. 

This is fine and will be time well invested.

> It's not for the faint of heart, and I do not have this kind of time anytime soon.

I was not expecting you to do this anytime soon. But I would appreciate some guidance if/when I find a qualified person to overhaul the build system. Thanks for all your help!

agoetz:
I moved the compiler flags into the root CMakeLists.txt, see PR #153. I will close this PR

PR #150, opened 2021-03-13
Adding API test into test suite

Madu86:
The changes in this PR should enable testing API through runtest script. The script will look for test-api, test-api.MPI, test-api.cuda and test-api.cuda.MPI executables in ${installfolder}/test/. I have updated make build system to direct the executables into correct location such that tests run without a problem. However, tests in cmake build are currently failing since the executables are not directed there. @multiplemonomials Can you make this change in the cmake build system?


agoetz:
@Madu86 This looks all good but I am not sure about the CMake installation. It should be specified here
https://github.com/merzlab/QUICK/blob/f4b704ffa41f2545c9027f5eeeeff4d13164f83d/src/CMakeLists.txt#L171
Problem is we don't have `${TESTDIR}` defined. This could be done in `cmake/InstallDirs.cmake` but my understanding is that within Amber this file will be replaced with Amber's version. You can add
```
install(TARGETS test-api DESTINATION "${CMAKE_INSTALL_POSTFIX}test" EXPORT QUICK)
```
etc for MPI, cuda, cuda.MPI targets, and it will put the api test executables into the test directory. I just tested this.

But I don't know what will happen when Quick is in Amber, because the test directory will be in a different location, and the tests are only optionally installed (afaik). The test directory is handled under `# Additional files` at the end of the root `CMakeLists.txt`. We might need an `IF(INSIDE_AMBER)` and choose a different destination? @multiplemonomials can chime in here.

Alternatively you could put the test-api executables into the bin folder. It's not nice to clutter the path with useless executables but it would work. And we could then fix the location later when we are not under time pressure.

Madu86:
@agoetz As far as Ive noted, the current cmake implementation doesn't generate api-test executables at all. If you can fix this, I can update the runtest script to look for executables in both bin and test directories.

agoetz:
@Madu86 the test-api executables are now installed in the bin folder. Please update your runtest script.

agoetz (changes requested):
Looks good. Only thing to change is to look for the test-api executables in the bin folder. This is where the cmake build puts them.

Madu86:
@agoetz I have updated runtest script to detect test-api executables from bin/test directories. The implementation seems to work fine. Please feel free to test and merge when CI tests are done.

agoetz (approved):
Looks good.

Issue #75, opened 2021-02-09
Possible bug related to XC gradients in serial and mpi versions


Madu86:
DFT gradients between cuda/cudampi and serial/mpi versions are significantly different for some molecules. For example, the following differences are observed for morphine B3LYP/6-31G test case. 

Serial/MPI version:

   39X    -3.1557002410     0.0103348688
   39Y    -2.6446002020     0.0002881292
   39Z     0.6672000509    -0.0120197547
   40X    -3.8483002939     0.0091268736
   40Y     1.0686000816     0.0105647101
   40Z     1.1031000842    -0.0082148827

CUDA/CUDAMPI version:

   39X    -3.1557002410     0.1090530496
   39Y    -2.6446002020     0.0338787728
   39Z     0.6672000509    -0.2400229409
   40X    -3.8483002939     0.0763664437
   40Y     1.0686000816     0.2128912033
   40Z     1.1031000842    -0.1546795715

This is possibly coming from the host code computing grid weight derivatives.

agoetz:
@madu Is this still relevant? I just executed `grad_daspi_meoh_b3lyp_6-31g.in` with `quick.cuda` and `mpirun -np 16 quick.MPI` and I get gradients and point charge gradients that differ by less than 1e-6. Please close this issue if you have no further concerns.

Madu86:
Probably we have fixed this with recent improvements. Can you please confirm with this input?

```
DFT BLYP BASIS=6-31G CUTOFF=1.0d-9 DENSERMS=1.0d-8 GRADIENT
  
 O         -1.99300       -0.47400        1.20050
 O         -2.78260       -2.51680       -0.22180
 O         -3.62200        1.84210        0.55790
 N          3.17770       -0.01010        0.44740
 C          0.36990       -0.71490        0.66160
 C          1.29130       -1.30460       -0.41700
 C          2.34410       -0.20230       -0.76400
 C         -0.93770       -1.47500        0.98480
 C          1.19880       -0.39070        1.92010
 C         -0.25160        0.53020        0.12050
 C          1.68870        1.10520       -1.32790
 C          2.43150        0.47540        1.61260
 C          0.33840        1.43060       -0.74800
 C         -1.36310       -2.50670       -0.09240
 C          0.50660       -1.81830       -1.59710
 C         -1.56720        0.63060        0.51360
 C         -0.71550       -2.34920       -1.44250
 C          4.37330        0.79160        0.20840
[...]

PR #151, opened 2021-03-13
Fix for issue #75

Madu86:
The changes in this PR should fix the bug reported in issue #75.


agoetz (approved):
Looks all good now. At first I thought we are getting wrong energies and gradients because I compared to Orca and this other code that shall not be named. But it turns out that you had accidentally used the libxc string for B3LYP instead of BLYP. So I changed this and now we can reproduce results, both with cuda (single GPU tested) and CPU code (MPI tested). I also updated the saved test output for the BLYP tests. I will merge this.

PR #153, opened 2021-03-14
Update CMake build system compiler flags - attempt 2

agoetz:
Update defaults to match the configure based build system. Implemented in root CMakeLists.txt to prevent that compiler options will be overwritten by Amber build system when building inside of Amber. This will have to be changed in the future to make Quick independent of any Amber compiler settings.
- disable warnings so users don't get shocked
- warnings can be enabled with -DWARNINGS=TRUE
- downgrade frome -O3 to -O2 so we won't have nasty surprises with the release on untested platforms
- remove duplicate -O2 flag
- update version number to 21.03
- disable PGI, CRAY, CLANG as these have not been tested or are not working
This supersedes PR #149 which will be closed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.9.0) # version 3.9.0 needed for FindMPI and FindOpenMP versions which provide imported targets
 
-project(quick LANGUAGES NONE VERSION 20.03)
+project(quick LANGUAGES NONE VERSION 21.03)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/quick-cmake)
 if(NOT INSIDE_AMBER)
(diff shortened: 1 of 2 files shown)


agoetz:
Tested to work as expected on chinotto (2x RTX2080 Ti). Tests (serial, MPI, cuda, cuda.MPI)  pass for
- gnu 4.8.5, openmpi 4.0.5, CUDA 11.0
- Intel 19.1.1.217, openmpi 4.0.5, CUDA 11.0

agoetz:
@multiplemonomials can you please take a quick look to see whether I overlooked anything here? I put all the required compiler flag definitions into the root CMakeLists.txt so we don't inherit anything from Amber. It works as expected for Quick standalone. I just want to make sure I did not miss anything that might then accidentally be carried over when building within Amber. Thanks!

multiplemonomials:
Okay, this looks like it should work

agoetz:
Thanks @multiplemonomials . I will merge this.

Issue #141, opened 2021-03-10
Check basis set files for supported elements


agoetz:
We have Iodine in the def2-SV(P) and def2-SVP basis sets. These have to be used with the corresponding 28-electron def2-ECPs. We do have ECP routines but I am not sure if those work. I have not tested. I also do not see ECP info in the basis set files. If we don't support ECPs, then we need to take out elements that require ECPs.

Madu86:
@agoetz I will check all basis sets today. We should remove elements that require ECPs from def2 basis sets and throw appropriate errors if users try to use them.

PR #159, opened 2021-03-14
Updated and new basis sets

Madu86:
In this PR, I have added new basis sets (6-311G, 6-311G*, 6-311G**, 6-31+G**, 6-31++G**, 6-311+G(2d,p), 6-311++G(2d,2p), PC-0, PC-1) and cleaned up the existing ones. I have also updated the full test suite. The improvements should fix issue #141.


agoetz (approved):
Looks good. The pc-0 basis sets matches Orca exactly for RHF calculation. The pc-1 basis set is very close, the difference must be from Cartesian (Quick) vs spherical harmonic (Orca). Orca cannot use Cartesian functions.

Just a note: B3LYP is not a good idea to test because one never knows whether the Gaussian version or another version (with different VWN functional) is implemented.

PR #163, opened 2021-03-15
Fix for current CI failure

Madu86:
The changes here should fix the current CI failure.


Madu86:
This looks good to me. Merging now..

Issue #164, opened 2021-03-15
Test script fails to find test-api executables with legacy build


Madu86:
If a prefix is not used during  configuration stage of legacy build, and the runtest script is executed as `./runtest {options}` rather than `make test/ make fulltest` , the script will not be able to find the test-api executables. This must be fixed ASAP.

PR #165, opened 2021-03-15
Fix for issue 164

Madu86:
The minor changes here should fix issue #164.

### DIFF
--- a/tools/runtest
+++ b/tools/runtest
@@ -662,7 +662,7 @@ if [ "$serial" = 'yes' -o "$mpi" = 'yes' -o "$cuda" = 'yes' -o "$cudampi" = 'yes
       echo  "Error: $qbindir/quick not found."
       exit 1
     fi
-    if ! ( [ -x "$testdir/test-api" -o -x "$qbindir/test-api" ] ); then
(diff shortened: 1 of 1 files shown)


tools/runtest
```diff
@@ -662,7 +662,7 @@ if [ "$serial" = 'yes' -o "$mpi" = 'yes' -o "$cuda" = 'yes' -o "$cudampi" = 'yes
       echo  "Error: $qbindir/quick not found."
       exit 1
     fi
-    if [ ! -x "$testdir/test-api" ] | [ ! -x "$qbindir/test-api" ]; then
```
agoetz, line 665: This works but why do you need the round brackets? I always have to look up these things for bash and/or sh but it works for both without the round brackets:
```
sh-3.2$ touch x1 x2
sh-3.2$ chmod u+x x2
sh-3.2$ if ! [ -x x1  -o -x x2 ]; then echo "not executable"; fi
sh-3.2$ if ! [ -x x1  -o -x x3 ]; then echo "not executable"; fi
not executable
sh-3.2$ 
```

agoetz (approved):
Looks good. Please merge once CI passes.

PR #166, opened 2021-03-16
Remove CUDA from CUDA_MPIV version

multiplemonomials:


agoetz:
I tested this PR on chinotto with gnu 8.3.1, openmpi 4.0.5, cuda 11.0. All tests from short list pass for all build targets (serial, MPI, cuda, cuda.MPI). Tested with 2x RTX2080 Ti cards.

@vwcruzeiro please check if this fixes the problem with sander.cuda.MPI. Thanks!

multiplemonomials:
OK @agoetz try it now, I think it's fixed.

agoetz (approved):
Looks good. I just tested this and can verify that all (short) tests pass, including the cuda.MPI API test. This should also fix the problem with the cuda.MPI version of sander (to be tested).

PR #181, opened 2021-03-20
 Update CUDA configurations in CMake build

agoetz:
- Enables CUDA Toolkit 11.1 and 11.2
    - Adds SM 8.6 for Cuda > 11.0
    - Drops SM 3.0 for Cuda >= 11.0 (also updated in configure script)
    - Enables manual selection of multiple QUICK_USER_ARCH flags (e.g. "volta turing")

### DIFF
--- a/configure
+++ b/configure
@@ -938,7 +938,7 @@ if [ "$cuda" = 'yes' ] || [ "$cudampi" = 'yes' ]; then
       else
 
         echo "Configuring for SM3.0, SM5.0, SM6.0, SM7.0, SM7.5, SM8.0"
-        cuda_arch_flags="$sm30flags $sm50flags $sm60flags $sm70flags $sm75flags $sm80flags"
+        cuda_arch_flags="$sm50flags $sm60flags $sm70flags $sm75flags $sm80flags"
         cuda_dc_flags='-Xptxas --disable-optimizer-constants'
       fi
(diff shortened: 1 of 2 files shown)


agoetz:
@Madu86 please take a look. The logic should replicate more or less your build flags. Let me know if you see an issue. I tried a few cmake3 configures and it did what I intended it to do. This will configure for many supported architectures if QUICK_USER_ARCH is empty. You can now set multiple architectures e.g. like `-DQUICK_USER_ARCH="volta turing"`. The implementation is not perfect because it will not flag an error if you have a typo in one requested architecture, e.g. `-DQUICK_USER_ARCH='volta turrrrrrring' will not flag as an error but just configure for volta.

All of this can be improved but I do not have time. (e.g. also would make more sense to call it QUICK_CUDA_ARCH).

Can you please test this? And merge if CI passes (should pass because does not affect CPU builds).

Madu86:
@agoetz I have tested this on MSU HPCC. It works well, however, we still have a problem with both build systems. If I have cuda/11.0 and want to compile code for K40/K80 (SM35 and SM37 are only deprecated in cuda/11.0) with `--arch` or `-DQUICK_USER_ARCH` options, both build systems will put SM30 into nvcc compiler flag list and the compilation will crash halfway through. Perhaps in future, we can introduce a `--device/-DDEVICE` flag where user can specify the card?

Madu86 (approved):
Looks good to me.

Madu86:
All tests passed. I am merging this PR.

Issue #180, opened 2021-03-20
Replace latin characters in QUICK output


Madu86:
Latin characters should be replaced with English in QUICK output and reference test output files as this can lead to test failures on different platforms.

PR #182, opened 2021-03-20
Fix for issue 180

Madu86:
In this PR I have changed a latin character in standard QUICK output and in reference test output files. This should fix issue #180.


agoetz (approved):
Thanks Madu. I further updated the citations with name fixes and adding Jamie.

I compiled the code, ran a test, and made sure that dacdif passes on the output file. I will merge this PR.

Issue #187, opened 2021-03-30
Reusing density matrix in geometry optimizations and MD


Madu86:
The density matrix is not reused during MD/geometry optimization calculations. This must be fixed.

agoetz:
This has been fixed for geometry optimizations and API with #188 right?

@Madu86 I think you can close this PR. Maybe open a PR to implement writing results like density matrix, perhaps MOs e.g. for visualization purposes or analysis with external programs, gradients etc; and reading density matrix as guess from the checkpoint file. This can then be used to reuse the density matrix in file based interface.

Madu86:
@agoetz Yes. I am closing this PR.

PR #188, opened 2021-03-30
Fix for issue 187

Madu86:
I have fixed issue #187 in this PR. Specifically, I have enabled reusing density matrix during MD and geometry optimization.


agoetz (approved):
Looks good.

PR #196, opened 2021-06-03
cew

tjgiese:
This is the version of quick that has been implemented into the "cew" branch of gitlab.ambermd.org.

The state of this code is such that it won't compile as a stand alone application; it links to libcew (within AmberTools) and makes subroutine calls to libcew in src/modules/quick_cew_module.F90.


Madu86:
Hi @vwcruzeiro, were you able to check this PR?

PR #207, opened 2021-09-08
Env variable for ASE inteface

TanemuraKiyoto:
I wrote an ASE interface for QUICK, which required an environmental variable to be defined. I modified the configure file to add the environmental variable.

### DIFF
--- a/configure
+++ b/configure
@@ -1331,7 +1331,7 @@ cat > "$quick_prefix/quick.rc" << EOF
 export QUICK_INSTALL=$quick_prefix
 export QUICK_BASIS=\$QUICK_INSTALL/basis
 export PATH=\$QUICK_INSTALL/bin:\$PATH
-export ASE_QUICK_COMMAND=$installer
+export ASE_QUICK=$installer
 EOF
 
 if [ "$shared" = 'yes' ]; then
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -1331,6 +1331,7 @@ cat > "$quick_prefix/quick.rc" << EOF
 export QUICK_INSTALL=$quick_prefix
 export QUICK_BASIS=\$QUICK_INSTALL/basis
 export PATH=\$QUICK_INSTALL/bin:\$PATH
+export ASE_QUICK=$installer
```
agoetz, line 1334: Kio, what are you trying to achieve here? The variable `$installer` is used in a loop and will contain the name of the executable that is built. After the loop it contains the last executable that was built (quick, quick.MPI, quick.cuda, or quick.cuda.MPI).

ASE should decide which executable to use and check if it is present in `$QUICK_INSTALL/bin`. In case of the MPI executables it will need to call them with `mpirun -np N ` prefix.

agoetz (changes requested):
I do not think the suggested solution is robust. See my comments for the change. Also, we will need to support this for the CMake build, which just copies `quick-cmake/quick.rc` into `$QUICK_INSTALL` (this is actually a more robust version of quick.rc than what the configure script generated).

agoetz:
Closing this PR. Finding the QUICK executable should be handled on ASE side.

PR #209, opened 2021-09-22
Minor changes required for getting cew to work properly

Madu86:
The changes in this PR should help to get cew to work properly in QUICK/AMBER QM/MM.


agoetz (approved):
Looks good. This seems to be all that is missing for cEw to work accurately.

PR #220, opened 2021-12-08
Improvements to OEI kernels

Madu86:
I have eliminated register spillage of OEI kernels and made some improvements. Please review and merge.


agoetz (approved):
Looks all good. I also compiled and tested on chinotto with GNU 8.3.1, OpenMPI 4.0.5 and CUDA 11.0 and ran the short test suite. Tests passed.

PR #221, opened 2021-12-10
Support for AMD GPUs

Madu86:
I have enabled support for AMD GPUs in this PR. On a platform with an AMD gpu and rocm-4.5 or above installed, one can configure serial and multi-GPU versions using legacy build as follows. 

./configure --hip gnu (gnu compiler assumed)
./configure --hipmpi gnu 

The above will generate make.hip.in and running `make` will compile the code for architecture of the available GPU resulting in quick.hip/ quick.hip.mpi executables. It is also possible to specify an architecture with --arch flag. For MI50 and MI100 this would be --arch gfx906 and --arch gfx908 respectively. I haven't tested on other architectures. 

The performance of current version of the code on AMD GPUs is ~1.5-2x slower than on a NVIDIA card with similar specs (based on MI100 vs V100 taxol B3LYP/6-31G** gradient time comparison). This is mainly due to the higher register spillage of device kernels on AMD cards and the use of host diagonalizer for matrix diagonalization. The rocSolver diagonalizer is currently inefficient: https://github.com/ROCmSoftwarePlatform/rocSOLVER/issues/353 and the corresponding calls to DSYEVD should be enabled once it is improved. I have already implemented the necessary driver.  

Things remaining to do: 
1. Fixing CMake build
2. Some cleanups (eg. use of common preprocessor flags in place CUDA and HIP specific flags and etc)
3. Enabling QM/MM support with AMBER
4. Testing on different AMD GPUs and updating configure script with arch flags and etc. 
5. Revisiting performance tuning once ERI code is refactored

Task 1 and 3 will require linking all objects with hipcc compiler since device objects need a special treatment: https://github.com/ROCm-Developer-Tools/HIP/issues/2406. We should keep this in mind.


Madu86:
In light of PR #259, this PR has become obsolete. Therefore I am closing this.

PR #228, opened 2022-03-08
Disable CUDA f functions in CMake by default

agoetz:
The f function code produces wrong numbers. Disable it by default.


Madu86:
All CUDA and CUDAMPI full tests pass. Merging this PR.

PR #229, opened 2022-03-08
Some minor updates to configure build system

Madu86:
In this PR, I have updated configure script to 1) handle latest CUDA versions, 2) disable compiling f functions in the ERI code by default.


agoetz:
@Madu86 There is no way to compile code with f functions now. Is this intentional? Otherwise add a flag `--withf` or `--yesf` that can be set to trigger build of f code paths.

Madu86:
@agoetz I have fixed this. Please review and merge.

agoetz (approved):
Looks good. I like `enablef`.

PR #230, opened 2022-03-10
Rolling back to old anlytical sswder code

Madu86:
In this PR, I have made necessary change to disable numerical grid weight derivatives and enable old analytical implementation.


agoetz (approved):
Looks good. Nice that this required just a small change.

PR #236, opened 2022-03-15
Energy print fix

Madu86:
In this PR, I have updated some printing statements in SCF and geometry optimization codes and also updated the reference output files.


Madu86:
@agoetz, All tests passed. Since thins PR contains some minor changes, I am will go ahead and merge.

PR #240, opened 2022-05-24
Random fixes

Madu86:
In this PR, I have updated default values of primitive cutoff and number of scf cycles to start the incremental Fock/KS build. I have also eliminated useless `acutoff` variable from quick_method_module. All tests passed.


agoetz (approved):
Looks good.

PR #245, opened 2022-06-20
added getsswanader within subs/sswder.f90

tjgiese:
This routine computes the analytic gradients of the ss dft weights. cew module has been changed to use this routine, but the actual DFT code hasn't been touched.


Madu86:
@tjgiese Your changes look good to me. Merging this PR.

PR #254, opened 2022-07-19
CUDA atomic add fix

Madu86:
I have updated cuda code to make use of legacy and intrinsic atomic operations depending on the hardware architecture. This should prevent the compilation failure on Kepler and Maxwell cards. Intrinsic atomics are now used by default and in order to use legacy atomics, all we need to do is passing USE_LEGACY_ATOMICS preprocessor flag. I have made appropriate changes to Make build but similar changes should be made to  CMake build.


Madu86:
@agoetz Ive updated cmake files in commit #[1c9d3d6]. (https://github.com/merzlab/QUICK/pull/254/commits/1c9d3d669a7ba11e4d453999e3f8594ceb5cd3ae)

Issue #251, opened 2022-07-13
How can MOLDEN and NBO files be generated using QUICK?


jimkress65:
How can MOLDEN and NBO files be generated using QUICK?

Madu86:
Hi @jimkress65, Thanks for reaching out. Unfortunately, this feature is currently not available in QUICK; but will be added down the line.

jimkress65:
Thank you for the reply.

 

You might want to consider its implementation since the properties calculated by QUICK are few.  With the MOLDEN file, many properties can be calculated using Multiwfn and similar programs.  Also, some QC packages, e.g. ORCA, can use MOLDEN files as data sources for MO interchange.

 

Jim Kress

 

James Kress Ph.D., President

The KressWorks® Institute 

An IRS Approved 501 (c)(3) Charitable, Nonprofit Corporation

“Engineering The Cure” ©

(248) 573-5499

 

Learn More and Donate At:

Website:  <http://www.kressworks.org/> http://www.kressworks.org

 

Confidentiality Notice | This e-mail message, including any attachments, is for the sole use of the intended recipient(s) and may contain confidential or proprietary information. Any unauthorized review, use, disclosure or distribution is prohibited. If you are not the intended recipient, immediately contact the sender by reply e-mail and destroy all copies of the original message.

jimkress65:
Thank you for the reply.

 

You might want to consider its implementation since the properties calculated by QUICK are few.  With the MOLDEN file, many properties can be calculated using Multiwfn and similar programs.  Also, some QC packages, e.g. ORCA, can use MOLDEN files as data sources for MO interchange.

 

Here’s a link to a MOLDEN file format specification:

 

https://www.theochem.ru.nl/molden/molden_format.html

 

Jim Kress

 

James Kress Ph.D., President

The KressWorks® Institute 

An IRS Approved 501 (c)(3) Charitable, Nonprofit Corporation

“Engineering The Cure” ©

(248) 573-5499

 

Learn More and Donate At:

Website:  <http://www.kressworks.org/> http://www.kressworks.org

 

Confidentiality Notice | This e-mail message, including any attachments, is for the sole use of the intended recipient(s) and may contain confidential or proprietary information. Any unauthorized review, use, disclosure or distribution is prohibited. If you are not the intended recipient, immediately contact the sender by reply e-mail and destroy all copies of the original message.

Madu86:
Thanks for pointing this out and sharing the link. We will add this feature in next release.

[...]

PR #256, opened 2022-07-22
Molden exporting enabled

Madu86:
In this PR, as requested in issue #251, I have enabled capability to export data from QUICK into molden format. Users can put `EXPORT=MOLDEN` in input file and this would result in input_name.molden file in the working directory. Ive followed this article: https://www.theochem.ru.nl/molden/molden_format.html for molden format, however, molden and other visualization software Ive tried dont read SCF/geometry convergence data. Interestingly, other QM codes (eg. MOLCAS, PSI4) do not write these sections into molden files. Not sure why.. Below are some orbital images generated using QUICK's molden files and JMOL. 

Screen Shot 2022-07-22 at 11 00 12 AM
Screen Shot 2022-07-22 at 11 00 33 AM
Screen Shot 2022-07-22 at 11 02 55 AM
Screen Shot 2022-07-22 at 11 03 09 AM


.gitignore
```diff
@@ -15,3 +15,4 @@ debug.*
 *~
 *.bak
 make.in
+quick.rc
```
agoetz, line 18: I think we need quick.rc, the one in `quick-cmake/`. Shouldn't ignore it.

src/modules/quick_method_module.f90
```diff
@@ -755,6 +756,13 @@ subroutine read_quick_method(self,keywd,ierr)
             if (index(keyWD,'COARSEINT').ne.0) self%coarse_cutoff=.true.
             if (index(keyWD,'TIGHTINT').ne.0) self%tight_cutoff=.true.
 
+            ! read and enable exporting
+            if (index(KeyWD,'EXPORT=') .ne. 0) then
+                tempstring=' '
+                call read(keyWD, 'EXPORT', tempstring)
+                if(index(tempstring,'MOLDEN') /= 0) write_molden=.true.
```
agoetz, line 763: Should we throw an error if there is a typo/unknown export format?

agoetz (changes requested):
Looks all good with one exception. When doing a geometry optimization, we should export the optimized geometry along with the MOs of the optimized geometry. Currently we are exporting the initial geometry.

.gitignore
Madu86: Fixed.

src/modules/quick_method_module.f90
Madu86: Fixed.

Madu86:
> Looks all good with one exception. When doing a geometry optimization, we should export the optimized geometry along with the MOs of the optimized geometry. Currently we are exporting the initial geometry.

Please see my original note. I have implemented a subroutine to write geometries along an optimization: https://github.com/Madu86/QUICK/blob/molden/src/modules/quick_molden_module.f90#L229-L262 but for some reason molden doesnt read them. Do you have an example molden file I can look at? The QC codes that Ive used in past (eg. MOLCAS) dump a molden file during each step of optimization or MD. A single file would contain geometry and orbitals corresponding to a given step.

agoetz:
I placed a PR to your molden branch. It fixes a few things
- correct geometry in [Atoms] block (geometry from last optimization step)
- only MOs from last geometry step in [MO] block (Molden displays MOs only for the geometry in the Atoms block)
- correct format in [GEOMETRIES] block
This now allows visualization of the optimization trajectory alongside with the energy plot from the [GEOCONV] block.
The only thing that does not work is display of data from the [SCFCONV] block. I tried to fix the format, but it still does not work. Not an issue imo.

@madu please check the PR in your repo and merge it if it looks good to you. I can then approve this PR.

Madu86:
@agoetz I tested and merged your PR. Feel free to merge this PR when tests are done.

agoetz (approved):
Looking all good now.

PR #259, opened 2022-09-27
Support for AMD GPUs

Madu86:
I have enabled support for AMD GPUs in this PR. On a platform with an AMD gpu and rocm-4.5 or above installed, one can configure serial and multi-GPU versions using legacy build as follows.

```
./configure --hip gnu (gnu compiler assumed)
./configure --hipmpi gnu
```

The above will generate make.hip.in and running make will compile the code for architecture of the available GPU resulting in quick.hip/ quick.hip.mpi executables. It is also possible to specify an architecture with --arch flag. For MI50 and MI100 this would be --arch gfx906 and --arch gfx908 respectively. I haven't tested on other architectures.

To configure and build QUICK using CMake build system, following steps can be followed.

```
cmake .. -DHIP=TRUE -DMPI=TRUE -DCMAKE_INSTALL_PREFIX=$(pwd)/../install -DCOMPILER=GNU
make -j16 install
```

The performance of current version of the code on AMD GPUs is ~1.5-2x slower than on a NVIDIA card with similar specs (based on MI100 vs V100 taxol B3LYP/6-31G** gradient time comparison). This is mainly due to the higher register spillage of device kernels on AMD cards and the use of host diagonalizer for matrix diagonalization. The rocSolver diagonalizer is currently inefficient: https://github.com/ROCmSoftwarePlatform/rocSOLVER/issues/353 and the corresponding calls to DSYEVD should be enabled once it is improved. I have already implemented the necessary driver.
Furthermore, double precision atomic add is surprisingly inefficient on AMD cards. Therefore, I have set the integer based legacy atomic add as default for now.


src/cuda/gpu_get2e_subs.h
```diff
@@ -663,19 +762,124 @@ __device__ __forceinline__ void iclass_spdf10
...
+/*
+#if defined int_sp
+    QUICKDouble store[STOREDIM*STOREDIM];   
+#endif
+*/
     /*
      Initial the neccessary element for
      */
     
+#if defined int_sp
```
agoetz, line 774: Why is this code replicated? It is identical for int_sp, int_spd, int_spdf, etc.

src/hip/gpu_get2e_subs_hrr.h
```diff
@@ -0,0 +1,1899 @@
...
+                         (int) LOC3(devTrans, LOC2(devSim.KLMN,0,KKK-1,3,devSim.nbasis) + LOC2(devSim.KLMN,0,LLL-1,3,devSim.nbasis), \
+                                    LOC2(devSim.KLMN,1,KKK-1,3,devSim.nbasis) + LOC2(devSim.KLMN,1,LLL-1,3,devSim.nbasis), \
+                                    LOC2(devSim.KLMN,2,KKK-1,3,devSim.nbasis) + LOC2(devSim.KLMN,2,LLL-1,3,devSim.nbasis), TRANSDIM, TRANSDIM, TRANSDIM)-1 , \
+                         STOREDIM, STOREDIM) * devSim.cons[III-1] * devSim.cons[JJJ-1] * devSim.cons[KKK-1] * devSim.cons[LLL-1];
+    
+    return Y;
+}
+
+
+__device__ __forceinline__ int lefthrr1(QUICKDouble RAx, QUICKDouble RAy, QUICKDouble RAz,
```
agoetz, line 754: Should we declare the variables that do not change with `const` type qualifier (e.g. `const QUICKDouble RAx, ` etc)? In principle it can help the compiler, although I do not know if it affects performance in this case. It's not consistent here with the corresponding file in the `cuda` directory. Neither of which use `const` for all functions.

src/libxc/hyb_gga_xc_wb97.c
```diff
@@ -15,7 +15,7 @@
  
 typedef struct { 
   double c_x[5], c_ss[5], c_ab[5]; 
-#if defined CUDA || defined CUDA_MPIV
```
agoetz, line 18: Why is cam_omega not used in CPU code?

src/libxc/hyb_gga_xc_wb97.c
```diff
@@ -24,7 +24,7 @@ static const gga_xc_wb97_params par_wb97 = {
   { 1.00000e+00,  1.13116e+00, -2.74915e+00,  1.20900e+01, -5.71642e+00}, 
   { 1.00000e+00, -2.55352e+00,  1.18926e+01, -2.69452e+01,  1.70927e+01}, 
   { 1.00000e+00,  3.99051e+00, -1.70066e+01,  1.07292e+00,  8.88211e+00} 
-#if defined CUDA || defined CUDA_MPIV
```
agoetz, line 27: Why do we add an extra parameter for GPU code?

agoetz (commented):
Madu, can you check the trivial points I mentioned in comments? After that I will merge and move to next PRs.

src/cuda/gpu_get2e_subs.h
Madu86: I am aware of this. It was passed down from Yipu. Not sure why. I will have this cleaned up in a separate PR. There are some other useless stuff as well.

src/hip/gpu_get2e_subs_hrr.h
Madu86: You are absolutely right. It helps, thats why I added them in _sp and _spd functions. The ones I didnt touch are related to f kernels. I will make relevant changes in upcoming high angular momentum integral PR. Because if we decide to combine the f kernels, some of these functions may not useful at all.

src/libxc/hyb_gga_xc_wb97.c
Madu86: Oh it should! Thanks for pointing this out. I think we never detected this since meta GGAs are disabled.

Madu86: In CPU code, cam_omega is set by initialize function into params struct: https://github.com/merzlab/QUICK/blob/0e1ecf816bc65aa3567bc3f5a47786b0155562cc/src/libxc/hyb_gga_xc_wb97.c#L60-L104. In CUDA version, we just set them directly inside functional structs.

src/libxc/hyb_gga_xc_wb97.c
Madu86: These are the values of cam_omega. See my previous comment.

agoetz: Ah ok. Thank you for explaining.

agoetz (approved):
Looking good. Very nice work!

PR #261, opened 2022-09-30
Changed the docs version to latest

Madu86:
I have updated the documentation links in README file from v22.3.0 to latest. We can change this for future releases but keep as latest for the master branch.


agoetz (approved):
Looks good. We just need to remember to change this to the release number whenever we create a tag for a release.

PR #263, opened 2022-10-04
Timer fix

Madu86:
In this PR, I have implemented a new timing scheme to measure wall time. Legacy timing scheme (CPU timing) is still available and in order to use this instead of new, `-DREPORT_CPU_TIME` flag should be passed into fortran flag list.


Madu86:
@agoetz Please make sure to merge this branch only after testing and merging AMD PR.

agoetz (approved):
Looking good.

PR #264, opened 2022-10-07
Support for MAGMA in AMD version

Madu86:
I have made necessary changes to make use of dsyevd diagonalizer form MAGMA library in AMD version. MAGMA must be installed on the target platform. Users will have to pass `-DMAGMA=TRUE` and set `-DMAGMA_ROOT=magma_path` (or set MAGMA_ROOT environment variable or have magma.so in LD_LIBRARY_PATH) when configuring cmake for HIP. 
Performance w.r.t. lapack dsyevd is system dependent. For taxol b3lyp/6-31G** example (1160 basis functions), lapack dsyevd diagonalization takes ~20 s for 11 SCF iterations. MAGMA dsyevd on MI100 takes ~8 s. 
I havent enabled magma support for CUDA version. This can be done in another PR if necessary. Please merge PR #259 and #263 before merging this into master.


agoetz (approved):
Tested on ACP and works fine. Great work.

I think we should print to the output which LA packages we are using (built-in, MKL, cuBLAS/cuSOLVER, rocBLAS/MAGMA). We can do this later.

PR #265, opened 2022-10-08
Updates to AMD device info being printed into QUICK output

Madu86:
I have updated some information being printed into QUICK output. These include correct AMD device architecture and MAGMA library version. Please merge this PR only after merging PRs #259, #263, #264.


agoetz (approved):
Looking good.

PR #269, opened 2022-11-15
Protect the loop exit condition with MAX_DII_CYCLE

q-posev:
With the current setup, the SCF loop is interrupted after 300 iterations even when `SCF=VALUE` key contains `VALUE` more than 300 (tried with 1000).

The 300 is the result of `MAX_DII_CYCLE_TIME*quick_method%maxdiisscf`, where `MAX_DII_CYCLE_TIME` hardcoded value is 30 and the default `maxdiisscf` value is 10.

This PR fixes this issue so that the current loop break condition is only checked when the total number of DIIS cycles (I  named it `MAX_DII_CYCLE`) is more than the max number of SCF cycles.

Alternatively, the control of the `MAX_DII_CYCLE_TIME` can be given to the user by supplying the corresponding input keyword. Let me know if this might be of interest.

**Note:** I have only patched the conventional SCF module. The same fix can be applied in `quick_uscf_module.f90` and `quick_sad_guess_module.f90` modules.

### DIFF
--- a/src/modules/quick_scf_module.f90
+++ b/src/modules/quick_scf_module.f90
@@ -180,7 +180,7 @@ contains
      logical :: diisdone = .false.  ! flag to indicate if diis is done
      logical :: deltaO   = .false.  ! delta Operator
      integer :: idiis = 0           ! diis iteration
-     integer :: IDIISfinal,iidiis,current_diis
+     integer :: IDIISfinal,iidiis,current_diis, MAX_DII_CYCLE
      integer :: lsolerr = 0
      integer :: IDIIS_Error_Start, IDIIS_Error_End
      double precision :: BIJ,DENSEJI,errormax,OJK,temp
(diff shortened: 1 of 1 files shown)


agoetz:
Thank you for noticing this issue with limiting SCF iterations by a hardcoded value.

I looked over the DIIS code. The exit condition you pointed out (`idiis.gt.MAX_DII_CYCLE_TIME*quick_method%maxdiisscf` is redundant in the current implementation. `idiis` is always equal `jscf`, hence it is sufficient to check for `jscf` as exit condition to limit the number of SCF iterations. You did this effectively by introducing the if-statement.

It is best to simply remove the check and get rid of `MAX_DII_CYCLE_TIME`. The user can choose the number of SCF iterations and size of DIIS subspace. There is no need to limit the number of DIIS cycles. I will fix it and create a PR and will leave this PR open until then. Thanks!

agoetz:
I created PR #278 which should fix this issue. Thanks again for your contribution and bringing this to our attention. If you still have problems please let us know and/or submit another solution. Thanks!

q-posev:
I agree, there is no need to limit the total number of DIIS cycles. Glad that this PR was useful 🙂

PR #270, opened 2022-11-18
Some improvements to AMD version

Madu86:
This PR contains following improvements. 

1. Capability to pass architecture flags in AMD version. For eg. QUICK can be configured for MI2XX using CMake build system as follows. 
`cmake .. -DHIP=TRUE -DMPI=TRUE -DCMAKE_INSTALL_PREFIX=$(pwd)/../install -DCOMPILER=GNU -DHIP_TOOLKIT_ROOT_DIR=$ROCM_PATH -DMAGMA=TRUE -DMAGMA_ROOT=~/magma-2.6.2/install-gfx90a -DQUICK_USER_ARCH=gfx90a `
Currently only gfx90a and gfx908 architecture flags are supported. The code will be configured and compiled for MI100 (gfx908) if no architecture is specified. 
2. Tested and set optimal kernel launching parameters for MI2XX. Enabled double precision atomic adds in MI2XX version.
3. Updated API to pass flag to reuse density matrix, integral cutoff, density matrix convergence threshold and XC cutoff.

Please merge this PR only after merging PRs #259 , #263, #264 and #265.


src/hip/gpu_getxc.cu
```diff
@@ -12,10 +12,10 @@ __constant__ gpu_simulation_type devSim_dft;
 static float totTime;
 #endif
 
-#include "gpu_getxc.h"
```
agoetz: What is the reason for renaming the file?

src/hip/gpu_get2e_subs.h
```diff
@@ -90,11 +90,15 @@ To understand the following comments better, please refer to Figure 2(b) and 2(d
  */
 #ifdef OSHELL
 #ifdef int_sp
-__global__ void 
-__launch_bounds__(HIP_SP_2E_THREADS_PER_BLOCK, 1) get_oshell_eri_kernel_sp()
+__global__ void
+__attribute__((amdgpu_waves_per_eu(HIP_SP_2E_WAVES_PER_CU,HIP_SPD_2E_WAVES_PER_CU)))
```
agoetz, line 94: Should this be twice `SP`? There is `SPD` as second argument to `amdgpu_waves_per_eu`.

src/modules/quick_api_module.f90
```diff
@@ -516,11 +519,11 @@ subroutine run_quick(self,ierr)
   SAFE_CALL(set_quick_molspecs(quick_api,ierr))
 
   ! start the timer for initial guess
-  call cpu_time(timer_begin%TIniGuess)
+  RECORD_TIME(timer_begin%TIniGuess)
 
   ! we will reuse density matrix for steps above 1. For the 1st step, we should
   ! read basis file and run SAD guess.
-  if(self%firstStep .and. self%reuse_dmx) then
+  if((self%firstStep .and. self%reuse_dmx) .or. .not. self%reuse_dmx) then
```
agoetz, line 526: Simpler and identical result
```
if(self%firstStep .or. (.not. self%reuse_dmx)) then
```
because we always want to do a guess if we are on the first step or if we do not want to reuse the density.

agoetz (commented):
Looks good. I tested compiled and tested the code on AMD ACP. Please see minor comments and then I will merge.

src/hip/gpu_getxc.cu
Madu86: According to my tests, functional specific global kernels (for XC potential and XC gradient) are significantly faster than generic kernels (i.e. single global kernel that can call many functional device kernels) on AMD cards. I was planning to have preprocessable version of gpu_getxc.h and renamed the existing version to gpu_getxc_generic.h.

src/hip/gpu_get2e_subs.h
Madu86: You are absolutely right. Good catch Andy!

src/modules/quick_api_module.f90
Madu86: Thanks for pointing this out!

Madu86:
@agoetz Ive fixed the two issues you highlighted. Please check. Thanks!

src/hip/gpu_getxc.cu
agoetz: Got it. Thanks.

agoetz (approved):
Looks good. I will merge this PR.

PR #278, opened 2023-02-12
AWG: Remove redundant hard-coded SCF DIIS exit condition

agoetz:
The exit condition `idiis.gt.MAX_DII_CYCLE_TIME*quick_method%maxdiisscf` is 1) redundant in the current implementation and 2) leads to a hard-coded maximum number of SCF iterations (DIIS subspace size times hardcoded number of maximum DIIS iterations). This is dumb, also because `idiis` is always equal `jscf`, hence it is sufficient to check for `jscf` as exit condition to limit the number of SCF iterations.

Here we remove the redundant condition and get rid of unnecessary `MAX_DII_CYCLE_TIME`. The user can choose the number of SCF iterations and size of DIIS subspace. There is no need to limit the number of DIIS cycles.

### DIFF
--- a/src/modules/quick_sad_guess_module.f90
+++ b/src/modules/quick_sad_guess_module.f90
@@ -855,7 +855,6 @@ contains
            if(verbose) write (ioutfile,'(" PERFORM FINAL NO INTERPOLATION ITERATION")')
            diisdone=.true.
         endif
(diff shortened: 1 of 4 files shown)


agoetz:
This also fixes the bug of hard coded max 300 SCF cycles (actually 30 times DIIS subspace dimension) mentioned in PR #269

Madu86 (approved):
Changes look good. Nice!

Madu86:
@agoetz Merging this PR.

PR #286, opened 2023-03-10
Fix Amber test script for HIP executables

agoetz:
Run longer tests (same as for CUDA)


Madu86 (approved):
This looks good. Thanks @agoetz. I will merge the PR when the CI tests are done.

PR #287, opened 2023-03-10
Increase keywd length

agoetz:
Match API keyword string length to Amber.
Increase from 200 to 256.
Longer is better to prevent truncation e.g. with long libxc functional names etc.


Madu86 (approved):
Changes look good to me. Merging this PR. Thanks!

PR #292, opened 2023-07-13
Support for aug-cc-pVDZ basis set

Madu86:
In this PR, I have added the aug-cc-pVDZ basis set. Please merge once all tests pass.


agoetz (approved):
Thanks for adding this basis set.

PR #295, opened 2023-09-08
Enabled F function trap

Madu86:
In this PR, I have re-enabled the trap for F functions.


agoetz (approved):
Looks good. We need to update the 23.08 release. Maybe call it QUICK 23.08a, make a new release with note that this is identical to QUICK 23.08 except for the trap that disables incomplete f function code, and update the comment on the 23.08 release that basis sets with f functions are not supported although the code does not quit.

PR #313, opened 2024-02-27
An update to initial information print

Madu86:
In this PR, I have updated the print statements that print version and author list into the output file. Specifically, I added pipe symbols into each line to prevent them from being considered for test vs reference output file comparisons in the test suite. This should fix QUICK test failures due to version mismatches in the AmberTools test suite.


agoetz (approved):
Looks good.

PR #314, opened 2024-02-28
Fix opt tests

agoetz:
The optimization tests were not working as intended. Only one geometry optimization test was checking the geometry and only some the dipole moment. The reason was that the awk command in the runtest script in function check_opt() was looking for "OPTIMIZED GEOMETRY INFORMATION" but only one geometry is actually fully optimized to save time. This test fixes this problem and adds new tests.
- Search for "GEOMETRY INFORMATION" in output, which is printed both for fully and partially optimized geometries
- Let merad test fully optimize
- Add DIPOLE keyword where missing
- Update output files accordingly
- Add caffeine B3LYP/3-21G geometry optimization (from Baker test set) so we also test a larger molecule (not in short CPU test list)
- Update some comments in test lists and runtest test description


akhilshajan:
I checked and it is now working properly all test cases. Thank you for the fix.

akhilshajan (approved):

Madu86 (approved):

PR #317, opened 2024-02-29
G function trap

agoetz:
1) Add trap to quit with error message if a user tries to use G functions.
2) Remove SAD guess that was made for basis sets / elements with G functions.


Madu86 (approved):
These changes look good to me.

PR #322, opened 2024-03-05
Fix for compile time error that occurs with intel compiler tool chain

Madu86:
In this PR, I have disabled aoint subroutines which leads to a compile-time error when using the intel compiler tool chain. Please check and merge.

### DIFF
--- a/src/modules/include/eri.fh
+++ b/src/modules/include/eri.fh
@@ -574,6 +574,7 @@ subroutine iclass_cshell(I,J,K,L,NNA,NNC,NNAB,NNCD)
    LLL2=quick_basis%ksumtype(LL)+NBL2
 
    if (quick_method%nodirect) then
+#ifdef COMPILE_AOINT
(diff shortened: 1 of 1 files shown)


agoetz (approved):
The code change looks ok. I tested this with GNU compilers and Intel compilers. It fixes the linker issues with Intel compilers.

This compiles on Expanse A100 GPU nodes with Intel 19.1.3.304 20200925 compilers, CUDA 11.2, and OpenMPI 4.1.3.

However, UDFT gradients are wrong with Intel compilers. Following tests fail
- grad_ohrad_b3lyp_631gss
- opt_merad_b3lyp_631gss

Instead of 5 alpha and 4 beta electrons, the output shows 2.5 alpha and 2.5 beta electrons in the DFT quadrature. This does not seem to affect UHF.

I will merge this but we need to investigate what is going on with UDFT with Intel compilers.

PR #334, opened 2024-03-12
Fix for issue 333

### LINKED ISSUE #333: Wrong D and F MO coefficient order in QUICK generated .molden files
The MO coefficient section in .molden files written by existing molden exporter in QUICK doesn't follow the correct format for D and F orbitals. The correct format is as follows (ref:  https://www.theochem.ru.nl/molden/molden_format.html).  

```
   5D: D 0, D+1, D-1, D+2, D-2
   6D: xx, yy, zz, xy, xz, yz

   7F: F 0, F+1, F-1, F+2, F-2, F+3, F-3
  10F: xxx, yyy, zzz, xyy, xxy, xxz, xzz, yzz, yyz, xyz
```
This must be fixed ASAP.

Madu86:
The enclosed changes resolve #333. Please check and merge.

### DIFF
--- a/src/modules/quick_molden_module.f90
+++ b/src/modules/quick_molden_module.f90
@@ -367,6 +367,7 @@ subroutine reorder_mo_coeffs(co, KLMN, nbasis, i, reord_mo_vec, ierr)
     j=1
     do while (j <= nbasis)
         reord_mo_vec(j)=co(j,i)
+        ! order d functions. xx is the first, followed by yy, zz, xy, xz, yz
         if(KLMN(1,j) .eq. 2) then
             reord_mo_vec(j+1)=co(j+2,i)
             reord_mo_vec(j+2)=co(j+5,i)
@@ -375,6 +376,21 @@ subroutine reorder_mo_coeffs(co, KLMN, nbasis, i, reord_mo_vec, ierr)
             reord_mo_vec(j+5)=co(j+4,i)
             j=j+5
         endif
+
+        ! order f functions. xxx is the first, followed by yyy, zzz, xyy, xxy, xxz, xzz, yzz, yyz, xyz
+        if(KLMN(1,j) .eq. 3) then
+            reord_mo_vec(j+1)=co(j+3,i) 
+            reord_mo_vec(j+2)=co(j+9,i) 
+            reord_mo_vec(j+3)=co(j+2,i) 
+            reord_mo_vec(j+4)=co(j+1,i) 
+            reord_mo_vec(j+5)=co(j+4,i)
+            reord_mo_vec(j+6)=co(j+7,i)
+            reord_mo_vec(j+7)=co(j+8,i)
+            reord_mo_vec(j+8)=co(j+6,i)
+            reord_mo_vec(j+9)=co(j+5,i)
+            j=j+9
+        endif
+
         j=j+1
     enddo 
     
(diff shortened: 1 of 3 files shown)


src/modules/quick_molden_module.f90
```diff
@@ -97,11 +100,11 @@ subroutine write_coordinates(self, ierr)
            ! we need to do this because optimizers may return the geometry
            ! for the next step which may be stored in xyz
            k = self%iexport_snapshot - 1
-           write(self%iMoldenFile, '(3(4x,F10.4))') (self%xyz_snapshots(j,i,k),j=1,3)
+           write(self%iMoldenFile, '(3(4x,F13.10))') (self%xyz_snapshots(j,i,k),j=1,3)
```
agoetz, line 103: This looks dangerous. F13.10 has field width 13 and 10 significant digits. Add one for the decimal point and one for  a minus sign, and this leaves only a single digit for the integer part of the coordinate.

I would use '3(2x,F20.10))' as format.

src/modules/quick_molden_module.f90
```diff
@@ -97,11 +100,11 @@ subroutine write_coordinates(self, ierr)
...
          else
            ! if it's a single point calculation we can use xyz
            ! we can't use xyz_snapshots because they have not been populated
-           write(self%iMoldenFile, '(3(4x,F10.4))') (xyz(j,i),j=1,3)
+           write(self%iMoldenFile, '(3(4x,F13.10))') (xyz(j,i),j=1,3)
```
agoetz, line 107: Same as above suggestion to use '3(2x,F20.10))' format.

src/modules/quick_molden_module.f90
```diff
@@ -189,16 +229,19 @@ subroutine write_mo(self, ierr)
         endif
 
         write(lbl1,'(I5)') i
-        write(self%iMoldenFile, '(A11)') "Sym= a"//trim(adjustl(lbl1))
-        write(self%iMoldenFile, '(2x, "Ene= ", E16.10)') quick_qm_struct%E(i)
+        write(self%iMoldenFile, '(A11)') "  Sym= a"//trim(adjustl(lbl1))
+        write(self%iMoldenFile, '("  Ene= ", F15.10)') quick_qm_struct%E(i)
```
agoetz, line 233: This does not allow for large energies. Use F20.20 or scientific notation.

src/modules/quick_molden_module.f90
```diff
@@ -212,20 +255,25 @@ subroutine write_mo(self, ierr)
             endif
     
             write(lbl1,'(I5)') i
-            write(self%iMoldenFile, '(A11)') "Sym= b"//trim(adjustl(lbl1))
-            write(self%iMoldenFile, '(2x, "Ene= ", E16.10)') quick_qm_struct%Eb(i)
+            write(self%iMoldenFile, '(A11)') "  Sym= b"//trim(adjustl(lbl1))
+            write(self%iMoldenFile, '("  Ene= ",F15.10)') quick_qm_struct%Eb(i)
```
agoetz, line 259: Same as above, allow larger total energy numbers (use F20.20 or scientific notation).

agoetz (changes requested):
Format specifiers need adjusting to allow larger numbers for coordinates and total energy,

Madu86:
@agoetz, Thanks for pointing these out. I have made changes. Please check.

PR #335, opened 2024-03-12
F trap for cEW enabled QM/MM

### LINKED ISSUE #329: Add trap for cEW with f functions
The current 3-center integral implementation does not support F functions. Therefore, cEW-enabled QM/MM MD simulations with high angular momentum basis functions are currently not possible. We have to add a trap for this.

Madu86:
I have implemented a trap for F functions if the user tries to run cEW-enabled QM/MM with high angular momentum basis sets. This resolves #329. Please test and merge.


agoetz (approved):
Looks good.

PR #339, opened 2024-03-12
Error trap for UHF/UDFT calculations with basis sets containing F functions and CUDA version

### LINKED ISSUE #330: Add error trap for UHF/UDFT with F functions
Current ERI implementation with F functions does not support UHF/UDFT calculations. We have to add an error trap for this.

Madu86:
The error trap implemented here should resolve #330.


agoetz:
I tried to resolve the conflict in the web editor but it does not save my changes. Giving up for today. @madu you can also resolve the conflict and merge.

agoetz:
Tested on Expanse A100 nodes with F functions enabled. All tests of full test suite pass for serial, mpi, cuda, cudampi.

PR #342, opened 2024-03-15
Fix for 318 add-more-supported-basis-sets

akhilshajan:
I have added new basis set and SAD guesses for each


agoetz (approved):
Looks good. I updated comments and removed SAD guess files for unsupported elements. I tested all basis sets with a B3LYP single point calculation on C2H6. Thanks!

PR #346, opened 2024-03-17
Support for legacy atomics in cuda f gradient kernels

Madu86:
In this PR, I have made changes to CUDA F gradient kernels to support legacy atomic add. Users won't be able to run F kernels on Kepler cards though. This is because we moved device pointers and some constants from constant memory into the shared memory. This improves the performance and the implementation we have in F gradient source files ( https://github.com/merzlab/QUICK/blob/326-support-for-legacy-atomics-in-cuda-f-gradient-kernels/src/cuda/gpu_get2e_grad_ffff.cu#L259-L505, https://github.com/merzlab/QUICK/blob/326-support-for-legacy-atomics-in-cuda-f-gradient-kernels/src/cuda/gpu_get2e_grad_ffff.cuh#L1695-L1730) is my experimental implementation. We should implement some templates, clean this up, and update other OEI/ERI/LRI kernels to use shared memory. Will be very useful for performance gain on AMD cards. 
I have also reimplemented the ERI sorting algorithm: https://github.com/merzlab/QUICK/blob/326-support-for-legacy-atomics-in-cuda-f-gradient-kernels/src/cuda/gpu_get2e_grad_ffff.cu#L167-L251. The old sorting algorithm (https://github.com/merzlab/QUICK/blob/326-support-for-legacy-atomics-in-cuda-f-gradient-kernels/src/cuda/gpu.cu#L612-L1137) has a bug and fails for sort high angular momentum ERIs properly. We should move the sorting algorithm I implemented in gpu_get2e_grad_ffff.cu into gpu.cu and do the sorting before the SCF.

### DIFF
--- a/src/cuda/gpu_common.h
+++ b/src/cuda/gpu_common.h
@@ -189,6 +189,11 @@ if (error != cudaSuccess && error != cudaErrorNotReady)\
 typedef float  QUICKSingle;
 #define QUICKULL \
 unsigned long long int
+#ifdef USE_LEGACY_ATOMICS
+#define QUICKAtomicType unsigned long long int
+#else
+#define QUICKAtomicType double
+#endif
 
 /* 
  ****************************************************************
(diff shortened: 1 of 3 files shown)


agoetz:
I do not see changes to the ERI sorting algorithm in this PR. @Madu86 is there a reason why you added the comment about the ERI sorting algorithm here? I assume we should open an issue for code cleanup?

agoetz (approved):
Looks good to me.
I tested on Expanse A100 cards with code generated with `cmake .. -DCMAKE_INSTALL_PREFIX=../install -DCOMPILER=GNU -DMPI=TRUE -DCUDA=TRUE -DQUICK_USER_ARCH=ampere -DUSE_LEGACY_ATOMICS=TRUE -DENABLEF=TRUE`. 
All tests pass

Issue #327, opened 2024-03-08
CI tests fail when compiled with OneAPI compilers


Madu86:
I've created a CI workflow file for OneAPI compilers: https://github.com/Madu86/QUICK/blob/intel-ci/.github/workflows/serial-intel.yml. We have two problems: 1) CMake doesn't recognize compiler names, 2) tests fail after compilation: https://github.com/Madu86/QUICK/actions/runs/8198338051/job/22421701533. I got around 1 by creating symlinks but we have to fix both.

agoetz:
Guoquan Chen from Intel has compiled AmberTools 22 (works also with AmberTools 23) with Intel oneAPI LLVM based compilers. All codes compile, including QUICK. See branch `amber20-sycl-cmake` in the Amber Gitlab repo. He implemented this without CMake smarts (I think starting with CMake 3.27 there is built-in oneAPI support, but we should not rely on users having very recent CMake versions). Compiler is called IntelLLVM (-DCOMPILER=IntelLLVM) to distinguish from old Intel compilers.

Scott Brozell is also working on getting Amber/AmberTools compiled with Intel LLVM compilers. There is an issue / discussion on Amber Gitlab.

Dave Case recently merged support into Amber master. Take a look how this works (compilers called IntelLLVM or ONEAPI).

ohearnk:
I pulled in the most recent build support in AmberTools in #348.  Using this, I was able to successfully build standalone QUICK on the High Performance Computing Center at MSU via the following:

```
module purge && module load OneAPI/2021.3.0 CMake/3.18.4
cd QUICK
mkdir build_mpi_broadwell_intel_oneapi2021.3.0 && cd build_mpi_broadwell_intel_oneapi2021.3.0
cmake .. -DCOMPILER=ONEAPI -DMPI=TRUE -DCUDA= -DENABLEF= -DCMAKE_INSTALL_PREFIX=${PWD}/../install_mpi_broadwell_intel_oneapi2021.3.0
```

That said, some gradient and optimization tests are failing for both the serial and MPI (with 1 proc) implementations:

```
======================== Test Summary ========================

mpi version:
    Energy tests: 0/87 (failed/total)
    Gradient tests: 3/71 (failed/total)
    Optimization tests: 4/10 (failed/total)
    API tests: 0/1 (failed/total)
    Total tests: 7/169 (failed/total)

==============================================================
```

ohearnk:
Full diffs for the MPI tests:

```
[...]

PR #348, opened 2024-03-18
Add build support for Intel OneAPI / LLVM compilers

ohearnk:
This PR adds support for building with Intel OneAPI compilers (LLVM-based) and resolves issues when building with Intel MPI libraries.

*Note*: the aforementioned supported added in this PR  comes from pulling inupdates from AmberTools build system.  As this is currently still actively developed, this may require further updates.

Related: #327


agoetz (approved):
Looks good.

PR #349, opened 2024-03-19
Fix for issue 337

### LINKED ISSUE #337: Unrestricted calculations give wrong results with Intel compiler
UDFT calculations give wrong results with Intel compiler. GNU compiler work.

Following tests fail
- `grad_ohrad_b3lyp_631gss`
- `opt_c2h5rad_b3lyp_631gss`
- `opt_c2hrad_uhf_ccpvdz`
- `opt_merad_b3lyp_631gss`

The output shows wrong number of electrons in the XC quadrature. See https://pastebin.com/Er2avNaE for `grad_ohrad_b3lyp_631gss` which shows 2.5 alpha and 2.5 beta electrons in the XC quadrature. Correct number of electrons is 5 alpha and 4 beta.

Both UHF and UDFT fail. 

Tested on SDSC Expanse A100 nodes with `Intel compilers 19.1.3.304 20200925` and `CUDA 11.2`. Modules loaded as follows
```
module reset
module load intel/19.1.3.304
module load cuda11.7/toolkit
module load openmpi/4.1.3
module load intel-mkl/2020.4.304
module load cmake/3.21.4
```

Madu86:
The enclosed change should resolve #337. It seems that ifort represent logical true using 255 rather than 1. As a result, when we pass the "is_oshell" flag value from Fortran to C code (https://github.com/merzlab/QUICK/blob/master/src/cuda/gpu.cu#L385-L402), the value in C code was not set properly, and closed shell kernel was invoked for open shell XC calculation (https://github.com/merzlab/QUICK/blob/master/src/cuda/gpu_getxc.cu#L98-L113). Please test and merge.

### DIFF
--- a/src/cuda/gpu.cu
+++ b/src/cuda/gpu.cu
@@ -398,7 +398,8 @@ extern "C" void gpu_upload_method_(int* quick_method, bool* is_oshell, double* h
 	gpu -> gpu_sim.hyb_coeff = *hyb_coeff;
     }
 
-    gpu -> gpu_sim.is_oshell = *is_oshell;
+    bool interpreted_is_oshell = *is_oshell != 0;
+    gpu -> gpu_sim.is_oshell = interpreted_is_oshell;
 }
 
 #ifdef CEW


agoetz (approved):
Works as intended.
CUDA version tested on Expanse A100 nodes, compiled with GNU and Intel compilers.

PR #352, opened 2024-03-21
AWG Updates to infos

### LINKED ISSUE #331: Update information
README.md and initialize.f90 (https://github.com/merzlab/QUICK/blob/master/src/initialize.f90#L71-L110) should be updated for QUICK-24.03. The citation should also be updated. But make sure that all QUICK and QM/MM tests pass after initialize.f90 update.

agoetz:
Updates README, program header, and CMake version info.
- Version is now 24.03
- Support f functions
- HIP currently disabled, available in 23.08 but not 24.03
- Supports macOS ARM
- Ask to cite geometry optimization paper

@Madu86 I do not know how to update links to documentation. Links are set to point to version 23.08. This needs to be updated.

Closes #331


Madu86 (approved):
All changes look good.

Issue #353, opened 2024-03-22
Update Amber test script/reference test output files


Madu86:
The tests on Jenkins failed due to an update we made to QUICK output. To fix this, we must either update reference test output files (https://github.com/merzlab/QUICK/tree/master/test/saved) or the Runtest amber script (https://github.com/merzlab/QUICK/blob/master/test/Run.tests.amber). Can you take care of this quickly? To reproduce the test failure, simply compile the code, source quick.rc from the installation directory, go to the tests directory and run `Run.tests.amber` script. 

Thanks!

Madu86:
I updated reference files with pipe symbols but optimization tests and F tests are still failing. This is because 1) someone removed max geom opt cycles and generated reference .out files., 2) a trap for F tests is not implemented in the above Runtest script. I implemented one in /tools/runtest script. The same has to be done here. @akhilshajan, @ohearnk Can you take care of these tasks and update 353-update-amber-test-scriptreference-test-output-files branch?

agoetz:
We should get rid of the `Run.tests.amber` script. It duplicates functionality of the `runtest` script. We should use the `runtest` script instead. When I use the `runtest` script with appropriate modification in the Makefile that runs AmberTools tests, all test pass in Amber Gitlab branch`quick-amber` (only tested for serial and MPI builds). 

This works because we do not do a diff on the entire file, but just on .ene, .grad, and .opt files that contain the relevant data. Furthermore, the runtest script has a trap for F tests.

Only a minor modification to the `runtest` script is required. I created PR #354 for this.

Two things need to be done in Amber Gitlab branch `quick-amber`, @Madu86 can you do that after incorporating the updated `runtest` script?
1) Modify the Amber CMake to copy the `runtest` script into `$AMBERHOME/AmberTools/src/quick` and do not copy `Run.tests.amber` (we should just remove it).
2) Modify `$AMBERHOME/AmberTools/test/Makefile` to call `runtest` with appropriate flags (e.g. `--serial`) instead of `Run.tests.amber`

The corresponding section in `$AMBERHOME/AmberTools/test/Makefile` should look as follows:
```
test.Quick:
        @(if [ "$(QUICK)" = 'yes' ]; then \
        cd ../src/quick && ./runtest --serial ;\
        fi ;\
        )

test.Quick.MPI:
        @(if [ "$(QUICK)" = 'yes' ]; then \
        cd ../src/quick && ./runtest --mpi ;\
        fi ;\
        )

[...]

PR #354, opened 2024-03-23
AWG Update runtest script for Amber

agoetz:
In Amber
- Assume this runtest script gets installed into $AMBERHOME/AmberTools/src/quick/
- While binaries are in $AMBERHOME/bin.

This update will make sure we find the binary folder. It also removes the Run.tests.amber script which is redundant and does not work as intended.


Madu86 (approved):

Madu86:
Thanks. This resolves #353.

PR #359, opened 2024-04-05
ioutfile intent in rather than intent out

tjgiese:
file unit dummy argument intent appears to be incorrect, which could result in using an uninitialized value


agoetz (approved):
Looks good. Thanks for fixing.

PR #380, opened 2024-10-02
Update Github action artifacts (v4)

ohearnk:
This MR updates Github action artifacts to v4 versions (v1 and v2 deprecated, then removed -- see [here](https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/)).


agoetz (approved):
Looks good.

agoetz:
While we are at it, we should update the build command. Gihub hosted runners now have 4 CPU cores. See

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

We currently use `make` and `make -j2`. 

We should update to `make -j4` or just use `make -j`, which should automatically build on all available CPU cores.

ohearnk:
@agoetz Regarding bumping up the number of cores used in Github runners from 2 to 4 -- doing this causing the MPI tests to fail on the first test (both configure/make and CMake builds).  Moreover, the error code returned from most of these failed tests from quick.MPI appears to be a value of 2 (i.e., the MPI rank first detected to be aborting).

Can you check if there is a settting in the merzlab account regarding the Github runner configuration?  My best guess is that a change might be required to reflect the hardware upgrades.

Other than that issue, this should be ready to merge.

merzlab:
@ohearnk There are no settings regarding the Github runner configuration. There is an option for Larger Github-hosted runners, but this is restricted to Team & Enterprise accounts. Another option is Self-hosted runners. These are virtual machines for GitHub Actions workflows that you manage and maintain outside of GitHub. We should look into this.

I checked for the CPU in the Github runner:
```
> Run lscpu
Architecture:                       x86_64
CPU op-mode(s):                     32-bit, 64-bit
Address sizes:                      48 bits physical, 48 bits virtual
Byte Order:                         Little Endian
CPU(s):                             4
On-line CPU(s) list:                0-3
Vendor ID:                          AuthenticAMD
Model name:                         AMD EPYC 7763 64-Core Processor
CPU family:                         25
Model:                              1
Thread(s) per core:                 2
Core(s) per socket:                 2
Socket(s):                          1
```

The information on the Github website is misleading. These are actually 4 threads but only 2 physical 4 CPU cores.

Let's leave build and MPI runs at 2 cores.

merzlab (approved):
Looks good. I will merge this PR.

PR #381, opened 2024-10-19
Github CI Workflow Optimizations

ohearnk:
This PR speeds up the serial CPU tests in the Github CI workflow by running those tests in parallel using GNU parallel (`parallel`) in the `runtest` script -- set the variable `PARALLEL_TEST_COUNT` before running `runtest`.  Specifically, use 2 cores for running the serial CPU tests for Github workflow runners.

Furthermore, the ability to run tests is parallel is supported for MPI, single GPU, and MPI+GPU QUICK versions. For MPI and MPI+X versions, set both `PARALLEL_TEST_COUNT` and `DO_PARALLEL`. For GPU versions, set both `PARALLEL_TEST_COUNT` and `CUDA_VISIBLE_DEVICES`/`HIP_VISIBLE_DEVICES`, where the number of available GPUs must be at least the number of parallel tests (i.e., at least 1 GPU per each GNU parallel job).

Additional notes/limitations:
- Exporting shell functions for parallel execution with `parallel` required using `bash`-style function exporting.  As such, the default shell used in `runtest` was changed to `bash`.  This should not be a problem with QUICK integration with Amber as many of the test scripts in Amber use `bash`.
-  `parallel`l has the ability to manage the output to STDOUT (see, e.g., `--group`, `--keep-order`).  This is used to keep the test output to STDOUT in-order by test ID.  However, `runtest` also duplicates output to `runtest.log`, which `parallel` cannot manage; as such, simple file locking with spin locks was used to avoid race conditions on writes to this file.  However, while the test outpus are grouped correctly by output per job, the order of job output is not enforced.  As such, the test output order will vary from run to run.  If this becomes problematic for functional reasons, this can be improved in the future.


tools/runtest
```diff
@@ -414,158 +248,393 @@ set_qexe_testlist(){
...
+  test_names=()
+  total_tests=0 
 
-  if [ "$test_ene" = 'yes' ]; then
-    tot_ene_tests=`awk '{print $1}' "$test_list"| grep "ene" | wc -l`
-    total_tests=$((total_tests+tot_ene_tests))
+  if [ "$1" = 'yes' ]; then
+    tot_ene_tests="${#tests[@]}"
+    tests+=(`grep "^ene_" "$test_list" | awk '{print $1}'`)
+    readarray -t -O "${#test_names[@]}" test_names <<< `grep "^ene_" "$test_list" | awk 'BEGIN { FPAT="([^#]+)"; OFS="\n"; } {print $2}'`
```
merzlab, line 307: `readarray` is not portable. Behavior changed with bash versions. 

It works as expected on Almalinux 9.4 with bash 
```
[agoetz@flamey install]$ cd $QUICK_INSTALL/test
[agoetz@flamey test]$ echo $BASH_VERSION
5.1.8(1)-release
[agoetz@flamey test]$ readarray -t foo <<<  `grep "^ene_" "testlist_short.txt" | awk 'BEGIN { FPAT="([^#]+)"; OFS="\n"; } {print $2}'`
agoetz@flamey test]$ echo ${#foo[@]}
11
[agoetz@flamey test]$ echo ${foo[0]}
PSB5: RHF energy test: s and p basis functions
```

But this happens on a CentOS 7.9 box with bash 4.2.46.
```
[agoetz@chinotto test]$ cd $QUICK_INSTALL/test
[agoetz@chinotto test]$ echo $BASH_VERSION
4.2.46(2)-release
[agoetz@chinotto test]$ readarray -t foo <<<  `grep "^ene_" "testlist_short.txt" | awk 'BEGIN { FPAT="([^#]+)"; OFS="\n"; } {print $2}'`
[agoetz@chinotto test]$ echo ${#foo[@]}
1
[agoetz@chinotto test]$ echo ${foo[0]}
PSB5: RHF energy test: s and p basis functions Acetone: RHF energy test: s and p basis functions PSB3: DFT energy test: s and p basis functions, BLYP functional PSB3: DFT energy test: s, p and d basis functions, BLYP functional PSB3: DFT energy test: s and p basis functions, native B3LYP functional PSB3: DFT energy test: s, p and d basis functions, native B3LYP functional PSB3: DFT energy test: s and p basis functions, libxc LDA functional PSB3: DFT energy test: s and p basis functions, libxc GGA functional PSB3: DFT energy test: s and p basis functions, libxc hybrid GGA functional Water-2 MP2 energy test: s and p basis functions Water-2 MP2 energy test: s, p and d basis functions
```

Following should be portable. Can you please check?
```
test_names+=(`grep "^ene_" "$test_list" | awk 'BEGIN { FPAT="([^#]+)"; OFS="\n"; } {print $2}'`)
```

agoetz: This enables running tests in serial (without GNU parallel). 

However, when requesting to run tests simultaneously leads to hanging:
```
[agoetz@chinotto install-pr381]$ export PARALLEL_TEST_COUNT=2
[agoetz@chinotto install-pr381]$ ./runtest --serial

Testing QUICK-24.03 serial version started on 10/19/2024 at 08:19:38 PM. 
```
No further output.

Two output files are created in `$QUICK_HOME/test/runs/serial/` corresponding to the first two tests. These tests ran and quit with an error because the input file was not found
```
| Error: Failed to open input file.

| Error Termination. Task Failed on: Sat Oct 19 20:19:38 2024
```
However, no further tests run, which is odd.

agoetz: GNU parallel version on chinotto (CentOS 7.9 box)
```
[agoetz@chinotto install-pr381]$ parallel --version
GNU parallel 20160222
```
Same behavior is expected on RHEL 7, of which there probably still are some around.

tools/runtest
```diff
@@ -909,90 +1020,97 @@ fi
...
+    tc_end="$tot_ene_tests"
+    if [ $has_parallel = 'no' ]; then
+      for tc in `seq $tc_start $tc_end`; do
+        run_energy_test "${tests[$((tc-1))]}" "${test_names[$((tc-1))]}" "$tc" "$total_tests" \
+		"$testdir" "$QUICK_HOME/.quick.${buildtype}.runtest.log"
+      done
+    elif [ $has_parallel = 'yes' ]; then
+      parallel --keep-order --group --jobs "$PARALLEL_TEST_COUNT" --will-cite run_energy_test {1} {2} {3} \
+	      "$total_tests" "$testdir" "$QUICK_HOME/.quick.${buildtype}.runtest.log" {%} \
+	      ::: "${tests[@]:$((tc_start-1)):$tc_end}" :::+ "${test_names[@]:$((tc_start-1)):$tc_end}" \
```
agoetz, line 1060: Older parallel versions do not provide `:::+`. If I change this into `:::` then the tests run (no hang) on chinotto (CentOS 7.9) but produce errors.

agoetz: Replacing `:::+` with `--xapply` in combination with `:::` works.

tools/runtest
agoetz: Actually, the issue is not with readline, but command substitution requires quoting the returned string, otherwise newline characters are stripped, e.g.
```
# This works
readarray -t foo <<< "`grep ...`"
```

tools/runtest
```diff
@@ -414,158 +248,393 @@ set_qexe_testlist(){
...
-
-    rm -f "$t.ref.out" "$t.in" ddtmp* *.ene *.grad *.dipole *.opt
 
+# function to clean up files from test
+#
+# arguments:
+#   $1: test name
+clean_up()
+{
+  rm -f "$1.ref.out" "$1.in" "ddtmp.$1*" "$1.*ene" "$1.*grad" "$1.*dipole" "$1.*opt"
```
agoetz, line 631: Globbing doesn't work in single or double quotes. Needs to be changed.

agoetz:
@ohearnk I fixed the following
- wrap command substitution in string to retain newline characters for readarray
- fix removing temporary files (remove strings to enable globbing)
- fix parsing of temporary log file .quick.${buildtype}.runtest.log for parallel run
  (test output is not any more in order; cannot expect test i+1 to follow in test i in output)

Still not working
- parallel runs with older GNU parallel versions that do not support `:::+`
- using `--xapply` in place of `:::+` does not work, it leads to quick execution failures

I have checked running serial and MPI tests (I have not tested GPU versions) both sequentially and in parallel for cases in which all tests pass and also with an artificially failing test (I modified one saved test output):
- Sequential and parallel execution both work on flamey (Almalinux 9.4, GNU parallel 20240922).
- Only sequential works on chinotto (CentOS 7.9, GNU parallel 20160222).

I don't know an easy fix for old GNU parallel versions and I don't think it's worth spending time on. I suggest to disable GNU parallel for older versions. Can you please implement a version check and quit if someone sets `PARALLEL_TEST_COUNT` but has on old version of GNU parallel that does not support `:::+`. I don't know which version starts supporting this. We could simply pick the lowest version of which we know it works. Thanks!

ohearnk:
@agoetz Thanks for the fixes.  The globbing issue was a mistakenly left in from some debugging on my end.

Regarding older GNU parallel versions not supporting the shorthand `:::+` for argument linking -- `--xapply` does work in my testing on chinotto.  My most recent commit makes this change.  Note that if testing from within the QUICK `$install_dir`, be sure to manually copied over the updated `runtest` script from `tools/runtest`.

The real reason for the test hangs on chinotto was that there was a bug with the sub-arrays passed to GNU parallel (array slicing was incorrect).  This should be fixed now.

agoetz:
Thanks @ohearnk I can confirm that serial and MPI tests now work as expected.

Sequential CUDA and CUDA-MPI tests also work.

There is still something wrong with selecting GPU IDs for parallel tests with CUDA:
```
[agoetz@chinotto install-pr381]$ export CUDA_VISIBLE_DEVICES=0,1,2
[agoetz@chinotto install-pr381]$ export PARALLEL_TEST_COUNT=3
[agoetz@chinotto install-pr381]$ ./runtest --cuda

Testing QUICK-24.03 cuda version started on 10/20/2024 at 01:57:20 PM. 

Error: invalid # GPUs detected in HIP_VISIBLE_DEVICES/CUDA_VISIBLE_DEVICES (# GPUs >= PARALLEL_TEST_COUNT).
Error: invalid # GPUs detected in HIP_VISIBLE_DEVICES/CUDA_VISIBLE_DEVICES (# GPUs >= PARALLEL_TEST_COUNT).
...
```

agoetz:
@ohearnk please add info on how to set `DO_PARALLEL`, `PARALLEL_TEST_COUNT`, and `CUDA_VISIBLE_DEVICES` to the help message that prints with `runtest --help`.

ohearnk:
@agoetz Thanks, this should be ready to merge now.

agoetz (approved):
Looking good now. I have tested
- serial: sequential/parallel
- MPI: sequential/parallel
- CUDA: sequential/parallel
- CUDA-MPI: sequential

I have not tested CUDA-MPI parallel because I tested on chinotto, which has only 3 GPUs.

PR #389, opened 2025-01-27
Fixes Segmentation error during  MOLDEN export on CPU run

hovo1990:
This update resolves a segmentation fault issue that occurred when running the application on the CPU and attempting to export data to MOLDEN format.


agoetz (approved):
Looks good. I also tested the code on a local machine (gcc 11.5.0, Open MPI 4.0.5, cuda 12.5).

PR #390, opened 2025-01-30
AWG Fix readline compatibility for older bash versions

agoetz:
Older bash versions do not support readline.
This includes bash shipping with MacOS (due to licensing).
This PR replaces readline with while loop in runtest script.
(code generated by ChatGPT and tested on MacOS 15.2 with bash 3.2.57 and Almalinx 9.3 with bash 5.1.8)


ohearnk (approved):
Looks good, and all testing in parallel on my end is passing.

PR #402, opened 2025-02-21
Fix Setting Compilation Flags for GitHub CI Workflow Tests

ohearnk:
This PR switches to using the compiler type keywords (GNU, CLANG, etc.) for the build systems in order to automatically set the compiler executables and compilation flags used (and thereby tests the common selections by end users).

The current code does manually set the compiler executables but neglects to set the compilation flags (which is a bug).  This fixes this issue.


agoetz (approved):
Looks good to me. Thanks for the updates.

PR #403, opened 2025-02-26
GitHub CI Workflow Enhancements Part 5

ohearnk:
~~NOTE: do not merge this until the 3 gradient tests failures described below are fixed.~~
NOTE: the branch in the PR is based off of the work in #402, so merge that PR first before merging this work.

This PR adds support for building QUICK with the Intel oneAPI compilers (CMake via the `-DCOMPILER=INTELLLVM` keyword).  Also Github CI tests for Intel oneAPI builds are added for the following runners:

CMake builds:
- Ubuntu 24.04 (x86_64): Intel oneAPI (v2024.2.1, v2025.0.1)

Notes / Limitations:
- Builds using the bundled BLAS/LAPACK code lead to a large number of test failures (~50%) but builds with Intel MKL (oneMKL) resolve these issues.  If we want to use and test the bundled BLAS/LAPACK Fortran codes, this will need to be investigated and fixed (possibly by updating this code -- see, e.g., #393).
- This work backports the compilation flags used by AmberTools for Intel oneAPI builds (speficially, not using the `-ipo` Fortran flag due to related issues with interprocedural optimization for libraries).

Closes #327.

PR #406, opened 2025-03-05
All types of whitespace characters used for delineation are not correctly parsed in input files

vtripath65:
When locating blocks in put file, previously a tab was not being considered as a space.

Now, if a line contains only tabs and/or spaces its considered an empty line.

The issues in the test set related to external charges with a tab at the end of the file as a newline is now fixed.


ohearnk:
To summarize, it was discovered by @vtripath65 that tab characters are present and not correctly parsed in the following inputs for gradient tests:

- test/grad_naip_meoh_libxc_pbeh_def2svp.in
- test/grad_naip_meoh_libxc_pbeh_sto3g.in
- test/grad_ncnaip_b3lyp_631gss.in
- test/grad_ncnaip_b3lyp_sto3g.in
- test/grad_rpsb_meoh_rhf_321g.in
- test/grad_rpsb_meoh_rhf_sto3g.in

This mistake leads to the number of number of external point charges being incorrect, which also leads to issues with point charge gradients (which were being computed for non-existent point charges). 

This also is the underlying bug which is causing the gradient test failures in #403.

PR #361, opened 2024-04-15
HIP and MPI+HIP updates

ohearnk:
This MR ports the updated CUDA and MPI+CUDA codes (including recently merged f-function optimizations) to HIP and MPI+HIP versions, respectively.  Also, this MR begins to unify the CUDA and HIP versions to simpily future GPU code maintainence.

Additional work also included in this PR concerns the following items:
- Integer-based atomic operation code paths (used in place of double precision atomics -mainly for older GPUs lacking hardware support [pre-pascal on NVIDIA GPUs])
- Wrappers around CUDA/HIP APIs (better error checking)
- HIP diagonalization (rocsolver re-enabled, CPU code disabled)
- Several bug fixed (memory leaks, initialization errors) 
- ROCm version detection and disabling HIP builds against known buggy ROCm versions (>= v5.4.3)

Limitations / Known Issues:
- f function code (CMake flag `-DENABLEF=TRUE`) does not compile on HIP versions due to resource limitations (stack frame size exceeded in device functions).  Errors are similar to the following (from a build on the AMD Accelerator Cloud platform):
```
error: stack frame size (143248) exceeds limit (131056) in function '_Z19getGrad_kernel_ffffPiPS_PdPS1_PP15HIP_vector_typeIiLj2EEPPhS7_S2_ii'
26 warnings and 1 error generated when compiling for gfx90a.
CMake Error at quick_hip_kernels_generated_gpu_get2e_grad_ffff.cu.o.Release.cmake:287 (message):
  Error generating file
  QUICK/build_ubuntu22_hip_gfx90a_rocm5.4.2_enablef_commit_0456378d/src/gpu/hip/CMakeFiles/quick_hip_kernels.dir//./quick_hip_kernels_generated_gpu_get2e_grad_ffff.cu.o

gmake[2]: *** [src/gpu/hip/CMakeFiles/quick_hip_kernels.dir/build.make:2080: src/gpu/hip/CMakeFiles/quick_hip_kernels.dir/quick_hip_kernels_generated_gpu_get2e_grad_ffff.cu.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:570: src/gpu/hip/CMakeFiles/quick_hip_kernels.dir/all] Error 2
gmake: *** [Makefile:156: all] Error 2
```

Closes #344.

PR #411, opened 2025-03-17
HIP and MPI+HIP Updates Part 2

ohearnk:
This PR backports data structure initialization fixes in PR #361 for the single GPU codepaths to the MPI+GPU codepaths.  Moreover, this PR unifies portions of the GPU and MPI+GPU initialization code to simplify future maintainence.

### DIFF
--- a/src/gpu/cuda/gpu.h
+++ b/src/gpu/cuda/gpu.h
@@ -1,112 +1,104 @@
 /*
- *  gpu_startup.h
- *  new_quick
- *
  *  Created by Yipu Miao on 4/20/11.
  *  Copyright 2011 University of Florida. All rights reserved.
- *
  */
 #if !defined(__QUICK_GPU_H_)
 #define __QUICK_GPU_H_
 
 #include "gpu_type.h"
 
-//Madu Manathunga 07/01/2019: Libxc header files
 #include "util.h"
 #include "../xc_redistribute.h"
 
 
-// device initial and shutdown operation
-extern "C" void gpu_set_device_(int* gpu_dev_id, int* ierr);
-extern "C" void gpu_startup_(int* ierr);
-extern "C" void gpu_init_(int* ierr);
-extern "C" void gpu_shutdown_(int* ierr);
-
-extern "C" void gpu_get_device_info_(int* gpu_dev_count, int* gpu_dev_id,int* gpu_dev_mem,
-        int* gpu_num_proc, double* gpu_core_freq,char* gpu_dev_name,
-        int* name_len, int* majorv, int* minorv, int* ierr);
+extern "C" void gpu_set_device_(int *, int *);
+extern "C" void gpu_new(
+#if defined(MPIV_GPU)
+        int,
+#endif
+        int *);
+extern "C" void gpu_init_device_(int *);
+extern "C" void gpu_delete_(int *);
+extern "C" void gpu_get_device_info_(int *, int *, int *,
+        int *, double *, char *, int *, int *, int *, int *);
 
 // molecule, basis sets, and some other information
-extern "C" void gpu_upload_method_(int* quick_method, bool* is_oshell, double* hyb_coeff);
-extern "C" void gpu_upload_atom_and_chg_(int* atom, QUICKDouble* atom_chg);
-extern "C" void gpu_upload_cutoff_(QUICKDouble* cutMatrix, QUICKDouble* integralCutoff,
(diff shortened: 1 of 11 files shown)


src/gpu/cuda/gpu.h
```diff
@@ -1,112 +1,104 @@
...
 
 #include "gpu_type.h"
 
-//Madu Manathunga 07/01/2019: Libxc header files
 #include "util.h"
 #include "../xc_redistribute.h"
 
 
-// device initial and shutdown operation
-extern "C" void gpu_set_device_(int* gpu_dev_id, int* ierr);
```
agoetz, line 20: In general it is recommended to keep variable names in header files for code readability and documentation. Why remove them here?

agoetz:
Note that for CUDA MPI tests with 2 tasks, DFT gradient calculations
- fail when running  on a single RTX 2080 Ti GPU
- fail or exit due to out of memory error when running on a single RTX 2080 GPU
The output files indicate an issue with XC routines:
```
 ALPHA ELECTRON DENSITY    =     0.0000000000
 BETA ELECTRON DENSITY     =     0.0000000000
```
On the RTX 2080, the program quits with an appropriate GPU out of memory error message (thanks to @ohearnk for implementing this feature!) for test `grad_psb3_b3lyp_631gss`
```
[ERROR] GPU error: cudaMemsetAsync failure
  [INFO] At line 70 in file /scratch/agoetz/QUICK/src/gpu/cuda/../gpu_get2e_getxc_drivers.h on MPI processor 1
  [INFO] Error code: 2
  [INFO] Error message: out of memory
```
This error message is unfortunately suppressed when running the test scripts (I ran the test manually to check the error message). If possible we should not suppress error messages to facilitate debugging in particular for end users.

We probably have hit an edge case where there is sufficient memory for ERIs but parts of the XC code memory handling does not use the new GPU wrappers, so the code continues and produces wrong XC results. We should open an issue for this.

agoetz:
The full GPU test suite passes for CUDA MPI with 2 tasks when using 2 RTX 2080 Ti GPUs.
(tested on chinotto with gcc 10.2.1, openmpi 4.0.5. and cuda 12.0)

So to me this PR looks good, except I am not sure whether we really should remove variable names from header files.

src/gpu/cuda/gpu.h
ohearnk: In most C/C++ codes which I've seen and worked on, it isn't common to use variable names in function interfaces for documentation -- other documentation mechanisms are employed, typically based on code comments (e.g., doxygen).  Moreover, for external interfaces, variable names bleed implementation details into places which should be implementation-agnostic (and thereby is also an additional places which needs to be updated if the underlying function code is modified).  This point about leaking implementation details is particularly relevent for properierty codes.

TL;DR -- it's a philosphocical software engineering choice.  Moreover, as these header files are internal interfaces in QUICK, only developers (i.e., us) will likely care.  If you feel strongly about documenting function interfaces in C/C++ code, we could also pursue using something like doxygen in place of this convention, for both internal and external interfaces.

ohearnk: As a data point, see the header files distributed with CUDA.

ohearnk:
> We probably have hit an edge case where there is sufficient memory for ERIs but parts of the XC code memory handling does not use the new GPU wrappers, so the code continues and produces wrong XC results. We should open an issue for this.
> 
> QUICK should use 1 MPI task per GPU. So this is in general not a problem. We should still open an issue.

@agoetz: I ported the GPU wrappers from the OEI+ERI codes to the XC code.  Can you pull in the new commit & build to see what  that shows for the single GPU RTX 2080 TI gradient test?  If it's anything interesting, please open an issue with further details.

agoetz:
@ohearnk I pulled in the changes with GPU wrappers in the xc code. Behavior did not change. I still see the same wrong results when runnnig two MPI tasks on a single RTX 2080 Ti card. Something else is going on.

agoetz:
MPI on two cards and single GPU version on one card passes tests. (I did not compile/test F functions and did not run the full test suite).

src/gpu/cuda/gpu.h
agoetz: OK. I am used to see variable names in function declarations, but I have not worked too much with C/C++. As far as I can tell this is also the case for CUDA headers (I just looked at $CUDA_HOME/include/ nvblas.h).

In any event, we don't have time now - this is a discussion for another time.

Issue #391, opened 2025-01-30
GNU compiler flags broken for MacOS 15 with Apple silicon


agoetz:
Compilation on MacOS 15 (Sequoia) on Apple silicon is currently broken.

MacOS 15 uses Xcode 16. GNU compiler installation via macports supports GCC versions 13 and 14.

We add `-mtune=native` to GNU compiler flags. However, this is only supported for GNU/Linux. On a Macbook Pro with M3 Pro CPU running MacOS 15.2 (Sequoia) with Xcode 16.2 and GCC 14.2.0 installed via macports this leads to following error
```
clang: error: unsupported argument 'apple-m3' to option '-mtune='
```

The solution is to either remove `-mtune=native` for GNU compilers on MacOS with Apple Silicon or figure out which flags actually work.

ohearnk:
More broadly speaking, I would advise that recent the Amber build system updates (which are effectively the upstream for the QUICK build system) need to be ported over.  This was what was done for the QUICK v24.03 release.  Doing this may resolve this issue automatically if the Amber devs have already sorted this issue out.

Additionally, I would suggest expanding the Github CI testing to include [MacOS runners.](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job).  This appears to be possible using [workflow job variations via matrix structures](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow).  While the MacOS runners appear to be currently limited to M1 CPUs, expanded testing may have caught this issue sooner (assuming that the Github runners eventually expant beyond M1 CPUs to include M3 CPUs).

ohearnk:
Further to the second point above, this may have also caught the bash issue in #390 .

ohearnk:
Upon further reflection, this is probably best addressed as a note under the [Known Limitations](https://quick-docs.readthedocs.io/en/latest/known-issues.html) section of the QUICK documentation.  Something like: if you compiler lacks support for your hardware via `-mtune=native`, then you need to set your own compilar optimization flags.

[...]

PR #412, opened 2025-03-18
QUICK 25.03 Build System Updates

ohearnk:
This PR backports recent AmberTools build system updates and bumps the supported CUDA versions (and GCC compatibilty checks) in light of recent release.

Notably, recent AmberTools changes do not address #391.  Hence, this still needs to be addressed.


agoetz (approved):
Looks good to me.

ohearnk:
Note: documentation for using MKL with QUICK will need to be updated to reflect that any extra CMake flag is now required to force the build system to link against MKL / oneMKL: `-DTRUST_SYSTEM_LIBS=TRUE`

Per Github CI tetss, the full set of MKL-specific flags for CMake builds now looks like the following:
```
cmake .. -DCOMPILER=INTELLLVM  -DMKL_HOME=$MKLROOT -DMKL_MULTI_THREADED=FALSE -DMKL_HOME=$MKLROOT -DMKL_MULTI_THREADED=FALSE -DTRUST_SYSTEM_LIBS=TRUE
```

PR #413, opened 2025-03-23
Fixes from AmberTools 2025 Integration

### LINKED ISSUE #391: GNU compiler flags broken for MacOS 15 with Apple silicon

ohearnk:
This PR backports fixes from integration testing with Amber/AmberTools 2025.  Specifically, it includes the following changes:
- expands Github CI tests to cover MPICH for MPI versions (since Amber CI MPI tests use MPICH)
- updates QUICK CUDA version and compatibilty checks
- cleans up additional code during debugging efforts with Amber CI tests (fixes DL-FIND build issues; fixes uninitialized data reported in MPI comms with Valgrind using the OpenMPI wrappers; silences GPU wrapper warnings in XC code for open shell DFT tests)
- fixes observed build failures with QUICK and MPICH for the legacy configure/Makefile build system by switching from using static to shared libraries (and adds a few cleanups and better diagnostic info for build steps in Github CI tests)
- backports libcew build fix (only use one copy for libcew in QUICK CMake builds for MPI targets) and test utility script updates from Amber (dacdif)
- fixes GCC build issues with recent Apple Arm CPUs (>= M3) due to improper support for `-mtune=native`
- adds checks to MPI+GPU to abort when running with # MPI procs > # GPUs (due to lacking insufficient checks of GPU resources) and updates the runtest script in such situations to log tests as skipped
- enhances the runtest script to log QUICK executable output from stdout/stderr to a separate verbose log file for further analysis upon test failures

Closes #391.


ohearnk:
@agoetz: I've updated the summary attached to this PR to better describe what was being changed during the debugging. I'd read the summary there rather than the noise in my previous few comments.

src/modules/quick_calculated_module.f90
```diff
@@ -318,13 +318,17 @@ subroutine reallocate_quick_qm_struct(self,ierr)
...
-     if(quick_molspec%nextatom .gt. 0) then
-       if(allocated(self%ptchg_gradient)) current_size= size(self%ptchg_gradient)
-       if(current_size /= quick_molspec%nextatom*3) then
+     if (quick_molspec%nextatom > 0) then
+       if (allocated(self%ptchg_gradient)) then
+         current_size = size(self%ptchg_gradient)
+       else
+         current_size = 0
+       endif
+       if (current_size < 3*quick_molspec%nextatom) then
```
agoetz, line 327: This guarantees that we have sufficient memory. But it could lead to issues because we may assume somewhere else in the code that this array has dimension 3*nextatom. I have to check the code.

src/modules/quick_molspec_module.f90
```diff
@@ -193,25 +193,34 @@ subroutine reallocate_quick_molspec(self,ierr)
...
-     if(self%nextatom .gt. 0) then
-       if(allocated(self%extchg)) current_size= size(self%extchg)
-       if(current_size /= self%nextatom) then
+     if(self%nextatom > 0) then
+       if(allocated(self%extchg)) then
+         current_size = size(self%extchg)
+       else
+         current_size = 0
+       endif
+       if(current_size < self%nextatom) then
```
agoetz, line 202: Need to check that dimension of these arrays can be larger than required where these arrays are used.

src/modules/quick_molspec_module.f90
```diff
@@ -193,25 +193,34 @@ subroutine reallocate_quick_molspec(self,ierr)
...
 
-     if(self%nextpoint .gt. 0) then
-      if(current_size /= self%nextpoint) then
+     if(self%nextpoint > 0) then
+       if(allocated(self%extpointxyz)) then
+         current_size = size(self%extpointxyz)
+       else
+         current_size = 0
+       endif
+      if(current_size < 3*self%nextpoint) then
```
agoetz, line 218: Need to check that dimension of these arrays can be larger than required where these arrays are used.

agoetz (approved):
Looks good!
Tested
- on chinotto (2x RTX 2080 Ti, 1x Titan V, gcc 10.2.1, openmpi 4.0.5, cuda 12.0)
- serial, mpi, cuda, cudampi tests
- PARALLEL_TEST_COUNT, DO_PARALLEL, and CUDA_VISIBLE_DEVICES set to multiple different values, including combinations that should not work, to make sure our error checking in the runtest script works as expected.

PR #415, opened 2025-04-05
Fixes from AmberTools 2025 Integration Part 2

ohearnk:
This PR fixes setting the Fortran source compiler optimization flags in CMake builds (due to an issue with case-senstive naming with COMPILE_LANAUGE in generator expressions [FORTRAN -> Fortran]).

Credit to @vtripath65 for noticing performance regressions in Fortran sources which led to this fix.


agoetz (approved):
Looks good. OPT_FFLAGS are now applied to Fortran sources. Tested on delorean and chinotto with GNU compilers, including cuda executables (single and multiple GPU) to make sure that opt flags work with CUDA.

PR #419, opened 2025-04-11
Fixes from AmberTools 2025 Integration Part 4

### LINKED ISSUE #418: macOS GNU compiler flags still broken
I think I understand better what happens on macOS. I am copying text from a discussion from MR 1743 on Amber Gitlab.

GCC on mac uses clang somewhere under the hood. GCC (versions 14.2 and 13.3 at least) have apple-m3 in their supported mtune list (not apple-m4). They pass it on to clang. But clang 16.0.0 (from Xcode 16.2) only knows apple-m2 and quits with an error. clang 17.0.6 (from Xcode 16.3) supports apple-m3 and gfortran works (gcc and g++ don't work at all with clang 17.0.6, which is a known issue).

We thus have
1. clang 17.0.6 (Xcode 16.3)
- gcc/gfortran (`-DCOMPILER=GNU`) is broken.
- clang/gfortran (`-DCOMPILER=CLANG`) works with `-mtune=native` both for clang and gfortran.
2. clang 16.0.0 (Xcode 16.2) and presumably older versions
- gcc/gfortran (`-DCOMPILER=GNU`) and clang/gfortran (`-DCOMPILER=CLANG`) work but require `-mtune=native` to be removed from all GNU compilers (gcc/g++/gfortran).

Hence what we need to do is either
- remove `-mtune=native` from all GNU compilers (gcc/g++/gfortran) when on an apple-m3
because a user might have any clang version installed.

Or, better, if we can check for the clang version,
- remove `-mtune=native` from all GNU compilers (gcc/g++/gfortran) when on an apple-m3 and clang version is lower than 17

ohearnk:
This PR attemps (yet again) to try and fix CMake builds on Apple ARM CPUs for GCC and mixed Clang+GCC builds.

GCC and Clang version checks added are based on web searches into when support was added into GCC and Clang for various Apple ARM CPU revisions (i.e., search for, e.g., LLVM on Github and look through the commits).

If this approach does not work, it is advisable to abandom this effort of using `-mtune=native` on Apple platforms (unless the performance gains can be demonstrated to merit further efforts).

Closes #418.


agoetz:
The logic for GNU compilers is wrong. It needs to check for compatibility of the GNU compiler and the underlying clang compiler. I fixed it and created a PR to your branch. Please check. It works as intended on my mac with M3 Pro CPU, gnu 14.2 and clang 17.0.0 compilers. I think it should work as intended also with other combinations of gnu and clang compilers. Thanks!

agoetz (changes requested):
Please check the PR I placed in your branch.

agoetz (approved):
Looks good.

PR #453, opened 2026-03-04
MPI support on the AMBER-QUICK API interface to run parallel free energy calculation

lizhen62017:
Hi Kurt, 

I have implemented a local patch to resolve this issue. https://github.com/merzlab/QUICK/issues/392

Sadly, it does not compile well on the standalone version of QUICK. Thank you for taking a look at it.

PR #473, opened 2026-04-06
MPI Communicator Refactoring (in support of expanded QM/MM API)

ohearnk:
This PR is a clean-up of the work in #453 (credit to @lizhen62017 for work there).  Specifically, this work is to support expanding the QM/MM API used by codes such as Amber / AmberTools via MPI communicator refactoring (for applications such as multiple parallel TI calculations).


src/modules/quick_api_module.f90
```diff
@@ -125,6 +125,10 @@ module quick_api_module
 
 ! allocates memory for a new quick_api_type variable
 subroutine new_quick_api_type(self, natoms, atomic_numbers, ierr)
+#ifdef MPIV
```
agoetz: We don't need the `mpi` and `quick_mpi_module` modules here.

src/modules/quick_api_module.f90
```diff
@@ -135,6 +139,10 @@ subroutine new_quick_api_type(self, natoms, atomic_numbers, ierr)
   integer :: atm_type_id(natoms)
   integer :: i, natm_type
 
+#ifdef MPIV
```
agoetz: This block is not required.

src/modules/quick_mpi_module.f90
```diff
@@ -21,8 +21,13 @@ module quick_mpi_module
 !  AUTHOR      : Yipu Miao
 !------------------------------------------------------------------------    
 
+!ZL2026
 #ifdef MPIV
+    use, intrinsic :: iso_c_binding
     use mpi
+    public :: quick_comm, is_master
+    integer :: quick_comm = MPI_COMM_WORLD  ! default
```
agoetz: How does this work? Doesn't this require `MPI_COMM_WORLD` to be known at compile time?

Shouldn't we initialize the communicator after calling `MPI_Init`?

src/modules/quick_api_module.f90
```diff
@@ -773,50 +766,46 @@ subroutine gpu_upload_molspecs(ierr)
...
   implicit none
 
-  integer, intent(in) :: mpi_rank, mpi_size
+  integer, intent(in) :: mpi_comm, mpi_rank, mpi_size
   integer, intent(out) :: ierr
 
   ! save information in quick_mpi module
-  mpirank    = mpi_rank
-  mpisize    = mpi_size
+  quick_comm = mpi_comm
```
agoetz: Should we do this instead? It would make sure messages by the calling program don't interfere.
```
call MPI_Comm_dup(mpi_comm, quick_comm, ierr)
```

Would need to be finalized with
```
if (quick_comm /= MPI_COMM_NULL) then
   call MPI_Comm_free(quick_comm, ierr)
end if
```

src/modules/quick_mpi_module.f90
```diff
@@ -21,8 +21,13 @@ module quick_mpi_module
...
+    logical, save :: is_master = .true.
```
agoetz: Are we using `is_master` anywhere?

src/modules/quick_scf_operator_module.f90
```diff
@@ -279,6 +280,7 @@ subroutine get_xc(deltaO)
      use xc_f90_lib_m
 #ifdef MPIV
      use mpi
+     use quick_mpi_module, only: quick_comm
```
agoetz: `quick_comm` is not used in subroutine `get_xc()`, the use statement is not required.

src/modules/quick_timer_module.f90
```diff
@@ -474,6 +475,7 @@ subroutine mpi_setup_timer
 
         use quick_mpi_module
         use mpi
+        use quick_mpi_module, only: quick_comm
```
agoetz: `quick_comm` is not used in subroutine `mpi_setup_timer()`. The use statement is not required.

src/mpi_setup.f90
```diff
@@ -16,12 +17,12 @@ subroutine initialize_quick_mpi()
 
       if(.not. libMPIMode) then
         call MPI_INIT(mpierror)
-        call MPI_COMM_RANK(MPI_COMM_WORLD,mpirank,mpierror)
-        call MPI_COMM_SIZE(MPI_COMM_WORLD,mpisize,mpierror)
+        call MPI_COMM_RANK(quick_comm,mpirank,mpierror)
```
agoetz: Should we not set `quick_comm = MPI_COMM_WORLD` here after the call to `MPI_INIT`? Or create a new communicator with the same group of processes:
```
call MPI_Comm_dup(MPI_COMM_WORLD, quick_comm, ierr)
```

src/mpi_setup.f90
```diff
@@ -50,15 +51,15 @@ subroutine mpi_setup_job(ierr)
     integer, intent(inout) :: ierr   
     
     call Broadcast(quick_method,ierr)
-    call MPI_BCAST(natom,1,mpi_integer,0,MPI_COMM_WORLD,mpierror)
-    call MPI_BCAST(quick_molspec%nextpoint,1,mpi_integer,0,MPI_COMM_WORLD,mpierror)
-    call MPI_BCAST(nbasis,1,mpi_integer,0,MPI_COMM_WORLD,mpierror)
+    call MPI_BCAST(natom,1,mpi_integer,0,quick_comm,mpierror)
```
agoetz: `quick_comm` is not defined in subroutine `mpi_stup_job()`. Not sure how this can work.

agoetz: Oh, I see, it's in `allmod`. At some point we should remove `allmod`.

agoetz:
Looks all good, just my minor comments and questions above.

In addition we need logic for handling assignment of hardware resources, in particular GPUs, when a calling program splits its communicator and calls QUICK as library for each sub-communicator. At present, multiple sub-communicators would use the same GPUs if they have visibility of the same hardware (e.g. tasks 0,1 and 2,3 on a node with 4 GPUs - QUICK would then use GPUs 0 and 1 for both sub-communicators).

ohearnk:
@agoetz - MPI+CUDA tests are now passing (tested 1, 2, and 4 GPUs on MSU HPCC, V100's).

This should be ready to merge.

ohearnk:
@agoetz - This now builds and passes multi-GPU tests on SDSC Chinotto.  It should now be ready for additional testing / review before approval for merging.

agoetz (commented):
Looks good. I have a few minor fixes. I will push these to this branch. We are running tests without/with f functions on chinotto, Expanse, and Cosmos. Once everything is finalized and passes, we can merge this and make the 26.03 release.

agoetz (commented):
I tested this PR so far on
1) Chinotto (RTX2080 Ti)
- 1, 2 GPUs
- no f functions
- GNU 10.2.1
- OpenMPI 4.0.5
- CUDA 12.0
2) Expanse (A100)
- 1, 2 GPUs
- ENABLEF=TRUE
- GNU 10.2.0
- OpenMPI 4.1.3
- CUDA 12.2.140

I still need to test on Cosmos.

agoetz:
We have two issues on Cosmos.

1. HIP code with f functions fails to build (ENABLEF=TRUE)
Compilation of `gpu_get2e_grad_ffff.cu` fails:
```
error: can't find enough VGPRs for wwm-regalloc                                                                             
1 error generated when compiling for gfx942.                                                                                
```
This is also the case for the current master branch (a3351e08). Hence not related to this PR.

2. libxc fails for HIP version
Code built without f functions (ENABLEF=FALSE).
Serial tests pass.
For HIP version, all tests that use libxc fail with
```
QUICK executable output (stdout/stderr):                                                                                    
[ERROR] GPU error: hipHostAlloc failure                                                                                     
  [INFO] At line 85 in file /cosmos/vast/scratch/agoetz/QUICK/src/gpu/hip/gpu_libxc.h                                       
  [INFO] Error code: 1                                                                                                      
  [INFO] Error message: invalid argument 
```
HIP tests pass in master (a3351e08), hence this is a bug introduced in the current PR that needs to be fixed.

agoetz:
Removing option `hipHostMallocNumaUser` from the `gpuHostAlloc()` call in `src/gpu/hip/gpu_libxc.h` fixes the problem.

However, the option `hipHostMallocNumaUser` does work in the master branch. I do not know which changes in this PR affect this option.

As a final resort we could remove it.

vtripath65 (approved):
Everything looks good. Merging the branch.

PR #474, opened 2026-04-08
Add Python interface (pyquick) via f2py

vtripath65:
## Summary

This PR adds `pyquick`, a Python interface to QUICK built with f2py, enabling
users to set up and run QUICK calculations directly from Python without writing
input files.

## New files

- `src/pyquick/pyquick.f90` — Fortran module compiled by f2py into a Python C
  extension. Exposes subroutines for setting the calculation type, basis set,
  method keywords, and molecular geometry, as well as running the SCF and
  retrieving results (total energy, MO energies, density matrix, Mulliken/Löwdin
  charges).
- `src/pyquick/__init__.py` — Pure-Python wrapper that imports the f2py extension,
  checks the `had_error` flag after every Fortran call and raises `RuntimeError`
  on failure, and provides the `PyQuick` class with per-instance state and a
  `copy()` method for creating independent conformations.
- `src/pyquick/CMakeLists.txt` — Build rules for the f2py extension. **Before
  attempting to compile, the build system explicitly checks that both Python and
  numpy (including f2py) are available** via `find_program` and
  `execute_process`. A clear `FATAL_ERROR` is emitted if either is missing,
  rather than failing silently at compile time.

## Build system changes

- `CMakeLists.txt` — adds `-DPYTHON=TRUE` option.
- `src/CMakeLists.txt` — adds `add_subdirectory(pyquick)` guarded by
  `-DPYTHON=TRUE`.
- `quick-cmake/quick.rc` — prepends `<prefix>/lib` to `PYTHONPATH` so that
  `import pyquick` works after `source quick.rc`.

## Build instructions

```bash
cmake .. -DCOMPILER=GNU -DENABLEF=FALSE -DPYTHON=TRUE \
         -DCMAKE_INSTALL_PREFIX=$PWD/../install
cmake --build . --parallel $(nproc)
cmake --install .
source ../install/quick.rc
```

## Usage example

```python
from pyquick import PyQuick

job = PyQuick()
job.set_calc('HF')
job.set_basis('STO-3G')
job.read_geom('''
    O  0.000  0.000  0.000
    H  0.757  0.586  0.000
    H -0.757  0.586  0.000
''')
job.run()
print(job.total_energy)

# Copy a job and modify independently
job2 = job.copy()
job2.read_geom('''
    O  0.100  0.000  0.000
    H  0.857  0.586  0.000
    H -0.657  0.586  0.000
''')
job2.run()
print(job2.total_energy)
```


agoetz:
Converted this to draft so we don't merge it before releasing QUICK-26.03

PR #475, opened 2026-04-15
HDF5 checkpoint/restart: per-step coordinate and density saving with intermediate geometry read support

vtripath65:
## Summary

- Renames the `read_coord` keyword to `readxyz` for clarity and consistency with the write-side keyword `chk_write_xyz`.
- Extends HDF5 checkpoint files to record geometry and density at every step of an optimization: coordinates are appended to an extendable `opt_traj` dataset each step, the `xyz` dataset always reflects the latest geometry, and the density matrix is written after every SCF cycle.
- Adds `CHK_READ_XYZ=<N>` support so a restart job can read the geometry from an arbitrary intermediate optimization step (step N) out of a previous run's `.dat` file.

## New / changed files

| File | Change |
|---|---|
| `src/modules/quick_io_module.f90` | `write_hdf5_real8_rank2`: delete-then-recreate to avoid crash on repeated calls; new `read_hdf5_opt_traj` subroutine; extendable `opt_traj` dataset write |
| `src/modules/quick_molspec_module.f90` | Remove unconditional `iattype` zeroing in `allocate_quick_molspec` |
| `src/modules/quick_method_module.f90` | Fix integer-as-logical print for `readxyz` |
| `src/modules/quick_optimizer_module.f90` | Per-step append to `opt_traj` + rewrite of `xyz` dataset |
| `src/dlfind/dl_find.f90` | Same per-step append for DL-FIND optimizer path |
| `src/main.f90` | Write `xyz` / `iattype` HDF5 datasets for SP, gradient, and opt paths |
| `src/getMol.f90` | Route `readxyz==0` vs `readxyz>0` to correct HDF5 read calls |
| `test/chk_acetate_intermediate_step5_xyz.in` | New write-job input (opt + `chk_write_xyz`) |
| `test/chk_acetate_intermediate_step5_xyz_2.in` | New read-job input (`chk_read_xyz=5`) |
| `test/testlist_full.txt`, `test/testlist_full_gpu.txt` | Add `chk_acetate_intermediate_step5_xyz` |
| `tools/runtest` | Add `_intermediate_step5_xyz` branch in `check_chk` (mutually exclusive `if`/`elif` with `_xyz$` branch) |


agoetz:
I see this in test outputs:
`environment: line 43: check_hdf5_error: command not found`

src/modules/quick_scf_module.f90
```diff
@@ -194,6 +194,9 @@ subroutine electdiis(jscf,ierr)
 #if defined(MPIV)
      use mpi
 #endif
+#if defined(RESTART_HDF5)
+    use quick_io_module, only: write_hdf5_real8_rank2
```
agoetz: I think we should not expose such low level functions. It would be better to keep these private in the `quick_io_module` and have higher-level functions that are exposed to the user of the module. For instance, `chk_write_den()`, which can then be called without cluttering the code with ifdefs.

src/modules/quick_scf_module.f90
```diff
@@ -707,6 +710,13 @@ subroutine electdiis(jscf,ierr)
         !--------------- END MPI/ALL NODES -------------------------------------
   
         if (master) then
+
+#if defined(RESTART_HDF5)
```
agoetz: Here we could call a routine `chk_write_den()` to replace all the code within this `#if defined(RESTART_HDF5)` block and also the `#ifdef USEDAT` block. This would keep the code easier to read and maintain.

src/getMol.f90
```diff
@@ -40,12 +40,25 @@ subroutine getMol(ierr)
 
       ! read xyz coordinates from the .in file 
       if (.not. isTemplate) then
-        if (quick_method%read_coord) then
+        if (quick_method%readxyz .ge. 0) then
 #if defined(RESTART_HDF5)
```
agoetz: This whole code block would also be better in a public subroutine like `chk_read_xyz()` in the `quick_io_module`.

src/main.f90
```diff
@@ -37,7 +37,8 @@ program quick
     use quick_molden_module, only : quick_molden, initializeExport, exportCoordinates, exportBasis, &
          exportMO, exportSCF, exportOPT
 #if defined(RESTART_HDF5)
-    use quick_io_module, only: write_hdf5_info, write_hdf5_int_rank1, write_hdf5_real8_rank2
+    use quick_io_module, only: write_hdf5_info, create_hdf5_extendable_real8_rank3, &
```
agoetz: Same here, I would not use these low-level functions (they should be `private` in the `quick_io_module`.

src/main.f90
```diff
@@ -315,6 +303,32 @@ program quick
         ! One electron properties (ESP, EField) 
         call compute_oeprop()
 
+#if defined(RESTART_HDF5)
```
agoetz: This block would be two calls to `chk_write_den()` and `chk_write_xyz()` with the logic to decide whether to write or not in those functions. Called only if master.

tools/runtest
```diff
@@ -830,7 +839,12 @@ check_chk()
     "$2/dacdif" -k -a '1.0e-8' "$1_2.ene" "$1.ene" | tee -a "$3"
   fi
 
-  if [ `echo "$1" | grep '_xyz' | wc -l` -eq 1 ]; then
+  if [ `echo "$1" | grep '_intermediate_step5_xyz' | wc -l` -eq 1 ]; then
```
agoetz, line 842: This is fragile, it seems to work only for a single test? If someone adds a different test or renames the test, this will break.

agoetz (commented):
Logic looks good to me. But I would prefer if we would not expose and use low level subroutines from the `quick_io_module`.

agoetz:
Note to please update keywords in documentation once this gets merged.

src/modules/quick_scf_module.f90
vtripath65: For every property add separate function? I do agree that the implementation is very low level. I think there should be an interface like chk_write(). The interace should handle which function to call.

src/modules/quick_scf_module.f90
vtripath65: Usage of inteface should help here too. I do not know about USEDAT block at this moment.

tools/runtest
vtripath65: This is a unique test. Only test where we are testing intermediate data instead of final. I need to think about how to incorporate this test into the general framework.

vtripath65: I think the frazileness will make sure only that test is being tested there. There is no other test like that.

vtripath65:
@agoetz This branch is ready to be reviewed again. I have created a separate pull request in QUICK-docs for the updated keywords.

src/modules/quick_sad_guess_module.f90
```diff
@@ -39,24 +39,31 @@ subroutine getmolsad(ierr)
      use quick_gridpoints_module
      use quick_files_module
      use quick_exception_module
+     use quick_io_module, only: chk_read
```
agoetz: This is not used here.

src/modules/quick_sad_guess_module.f90
```diff
@@ -39,24 +39,31 @@ subroutine getmolsad(ierr)
...
      character(len=80) :: keywd
      character(len=20) :: tempstring
      character(len=340) :: sadfile
-     integer natomsaved
+     integer :: natomsaved
      type(quick_method_type) quick_method_save
      type(quick_molspec_type) quick_molspec_save
      integer, intent(inout) :: ierr
      logical :: use_cew_save
+     integer :: istart, ifinal, ibasisstart, ibasisend
```
agoetz: These variables don't seem to be used.

src/modules/quick_io_module.f90
```diff
@@ -543,8 +594,738 @@ subroutine write_hdf5_real8_rank2(Array, length1, length2, datasetname)
...
+    double precision, dimension(n1, n2), intent(out) :: array
+    integer,                             intent(out) :: fail
+
+    integer   :: i, j, k, l, num
+    character :: kline*40, ktype*2, line*41
+
+    fail = 0
+#if defined(RESTART_HDF5)
+    call read_hdf5_real8_rank2(key, (/1,1/), (/n1,n2/), array)
+    fail = 1
```
agoetz: Why are we setting fail to 1? If we are not returning failure code from read_hdf5_real8_rank2, it seems we should leave it at 0. This is done in several places.

src/modules/quick_io_module.f90
```diff
@@ -543,8 +594,738 @@ subroutine write_hdf5_real8_rank2(Array, length1, length2, datasetname)
...
+  subroutine chk_append_opt_traj(natom, xyz, fail)
+    implicit none
+    integer,                          intent(in)  :: natom
+    double precision, dimension(3,natom), intent(in)  :: xyz
+    integer,                          intent(out) :: fail
+
+    fail = 1
+#if defined(RESTART_HDF5)
+    call append_hdf5_extendable_real8_rank3('opt_traj', 3, natom, xyz)
+    call write_hdf5_real8_rank2(xyz, 3, natom, 'xyz')
```
agoetz, line 1288: Is there a reason to use different order for arguments?
`('foo', 3, natom, xyz)` <-> `(xyz, 3, natom, 'bar')`

src/modules/quick_io_module.f90
```diff
@@ -543,8 +594,738 @@ subroutine write_hdf5_real8_rank2(Array, length1, length2, datasetname)
...
+    write(chk_unit) ((array(i,j), i=1,n1), j=1,n2)
+    fail = 1
+#endif
+  end subroutine chk_write_real8_rank2
+
+  !---------------------------------------------------------------------!
+  ! chk_read_int_scalar: read one integer scalar from the checkpoint.   !
+  ! non-HDF5: opens the binary file, rewinds, searches by key, closes.  !
+  ! HDF5: maps key to the appropriate 'molinfo' element index:          !
+  !   'natom'  -> molinfo[1]                                            !
```
agoetz, line 1110: Why are we doing this? This prevents reading a different scalar key. If someone tries to use it to store an integer under a different key it will fail.

src/main.f90
```diff
@@ -221,38 +220,17 @@ program quick
...
-            if (quick_method%writeden) then
-              call write_hdf5_real8_rank2(quick_qm_struct%dense, nbasis, nbasis, 'dense')
-              if (quick_method%UNRST) then
-                call write_hdf5_real8_rank2(quick_qm_struct%denseb, nbasis, nbasis, 'denseb')  
-              end if
+        if (master .and. quick_method%writechk) then
+            call chk_write('xyz', 3, natom, quick_molspec%xyz, fail)
+            call chk_write('iattype', natom, quick_molspec%iattype, fail)
+#if !defined(RESTART_HDF5)
+            call chk_write('dense', nbasis, nbasis, quick_qm_struct%dense, fail)
```
agoetz: Why execute this code block only  if the code was not compiled with HDF5 support? Shouldn't the logic for writing density to the DATA file and closing the DATA file be the same?

src/main.f90
```diff
@@ -272,37 +250,16 @@ program quick
...
-            if (quick_method%writexyz) then
-              call write_hdf5_int_rank1(quick_molspec%iattype, natom, 'iattype')
-              call write_hdf5_real8_rank2(quick_molspec%xyz, 3, natom, 'xyz')
-            end if
-            if (quick_method%writeden) then
-              call write_hdf5_real8_rank2(quick_qm_struct%dense, nbasis, nbasis, 'dense')
-              if (quick_method%UNRST) then
-                call write_hdf5_real8_rank2(quick_qm_struct%denseb, nbasis, nbasis, 'denseb')  
-              end if
+#if !defined(RESTART_HDF5)
```
agoetz, line 253: Why have this code block only if not compiled with HDF5 support?

src/main.f90
```diff
@@ -315,6 +272,18 @@ program quick
         ! One electron properties (ESP, EField) 
         call compute_oeprop()
 
+        if (master .and. quick_method%writechk) then
+            call chk_write('xyz', 3, natom, quick_molspec%xyz, fail)
+            call chk_write('iattype', natom, quick_molspec%iattype, fail)
+#if !defined(RESTART_HDF5)
```
agoetz, line 278: Why this conditional?

src/modules/quick_sad_guess_module.f90
vtripath65: Removed

src/modules/quick_sad_guess_module.f90
vtripath65: Removed the declaration and unused initializations.

src/modules/quick_io_module.f90
vtripath65: I have fixed this. I think the subroutine append_hdf5_extendable_real8_rank3 was called differently because it was written later along with create_hdf5_extendable_real8_rank3. It is a special subroutine for optimization trajectory only instead of other general purpose routines.

src/modules/quick_io_module.f90
vtripath65: We are currently reading only natom and nbasis. I will leave it the way it is at this moment. We will circle back to this later when we will expand the scope of restart and checkpoint capabilities.

src/main.f90
vtripath65: In case of HDF5, the density is already updated in the SCF module.

src/main.f90
vtripath65: In case of HDF5 support, the latest xyz and density is already present in the data file.

src/main.f90
vtripath65: same answer as previous. But, I can actually remove the conditional with no effect.

src/modules/quick_io_module.f90
vtripath65: We are not checking the value of fail. If HDF5 is defined, our error checking is robust after every step. But, otherwise there is no error checking.

vtripath65: For HDF5, we need don't need fail. For non-HDF5 case, the error checking needs to be added to the individual subroutines. I think we should just remove "fail" from everywhere.

agoetz: OK, it looks like convention in the old subroutines is
- success: `fail = 1`
- error: `fail = 0`
and the subroutines assume failure until I/O (read or write) was successful.
We can intercept the value of fail in the public wrapper functions and stop the program if there was an error. We shouldn't pass `fail` out of the public wrapper functions to be consistent with the HDF5 functions, which stop internally if something goes wrong but don't pass an error code.
The public wrapper functions then would look something like:
```
#if !defined(HDF5_RESTART)
  integer :: fail
#endif

#if defined (HDF5_RESTART)
  call read_hdf5_...() ! errors handled internally
#else
  call read_...(..., fail)
  if (fail == 0) ! error was encountered
    call PrtErr(OUTFILEHANDLE, 'I/O Error (chk_write)')
    call quick_exit(OUTFILEHANDLE, 1)
  end if
#endif
```

agoetz (approved):
Looks good to me. All important issues have been resolved. Further improvements can be done after release as necessary.

PR #476, opened 2026-04-17
MVP CI/CD Implementation for QUICK CUDA (without f functions) Build and Test Workflow

jxnlee:
Ensures that on every push and pull request, automate tests on a runner on NRP Nautilus.

Cannot build with f functions yet since the runner fails during build. Working on reconfiguring the QUICK runner to be able to handle building with f functions. Will introduce as a separate workflow.

Currently can only build/test on NVIDIA-A10s. Will introduce compatibility for other architectures next to ensure more availability.


agoetz (approved):
LGTM.

Future improvements:
- For clarity, we may want to go back to label it as cmake test, even if we only ever test the cmake build system for GPUs. 
- GPU architecture type is hardcoded as ampere. We may want to do this via a variable so it can be changed more easily.

PR #477, opened 2026-04-20
 MVP CI/CD Implementation for QUICK CUDA (with f functions) Build and Test Workflow

jxnlee:
This workflow is for QUICK build and test CUDA version with f functions enabled.

This version can take a long time and thus may (at times) timeout/disconnect (ie. because of how long it takes, it can accidentally disonnect from server node). The time for build can vary / be unpredictable. The maximum timeout duration has been set to 48 hours to best ensure that the job can run to completion, however this does not guarantee it won't lose connection. The default build and test workflow should still be without f functions enabled, which should run automatically (#476 ) with this (enablef) workflow being ran only if needed.

Because of the memory demands of this workflow, the runner requests 48GB of memory. Since the workflow without f functions could pass with just 8GB of memory it may be optimal to have separate runners but for now we just have one runner  requesting 48GB of memory.

The workflow can only be triggered manually through the run workflow button.

### DIFF
--- a/.github/workflows/build_test_cuda.yml
+++ b/.github/workflows/build_test_cuda.yml
@@ -11,11 +11,13 @@ on:
       - '**.md'
       - '.github/workflows/build_test_serial.yml'
       - '.github/workflows/build_test_mpi.yml'
+      - '.github/workflows/build_test_cuda_enablef.yml'
   pull_request:
     paths-ignore:
       - '**.md'
       - '.github/workflows/build_test_serial.yml'
       - '.github/workflows/build_test_mpi.yml'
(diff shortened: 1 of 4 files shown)


agoetz:
@jxnlee there was a modify/delete merge conflict because file `.github/workflows/build_test_cuda_enablef.yml` was already in the history  but deleted in master (probably from merge of PR #476 ). When doing the merge I messed something up.

I rescued the file from this branch and readded from a clean branch. So there is only one commit for this file now. I added a commit on top to exclude changes to this file from triggering serial/mpi/cuda tests.

agoetz (approved):
LGTM

PR #478, opened 2026-04-21
CMake Updates for AmberTools and CUDA v13.2.1

ohearnk:
This PR contains CMake updates backported from recent AmberTools 2026 (rc6), along with improved version check support for CUDA v13.2.x builds.


agoetz (approved):
Looks good. Thanks!

PR #479, opened 2026-04-22
Bump NVHPC SDK version in Github CI tests

ohearnk:
This PR updates the NVIDIA HPC SDK version installed and used for Github CI tests (v25.1 => v26.3).


agoetz:
Looks like there are problems with new PGI compilers. This requires some debugging.

PR #482, opened 2026-04-23
convergence threshold of esp calculations are loosened to 4.0e-8

vtripath65:
…better transferability across architecture.


agoetz (approved):
Looks good. Thanks for updating the thresholds.

PR #490, opened 2026-07-11
Refactor - Remove SCF from gradient calculation

agoetz:
Remove computation of energy (SCF) from getGradient() function.

Separating tasks cleanly enables for instance to read results from a prior SCF calculation, then skip the SCF and compute the gradient.

This is just a quick refactor start. There is still code duplication (e.g. GPU uploads etc) in the main program and scf/grad/optimizer code paths. This should be cleaned up and made more modular.

### DIFF
--- a/src/main.f90
+++ b/src/main.f90
@@ -214,23 +214,26 @@
     timer_cumer%TIniGuess=timer_cumer%TIniGuess+timer_end%TIniGuess-timer_begin%TIniGuess &
                           -(timer_end%T2elb-timer_begin%T2elb)
 
-    if (.not.quick_method%opt .and. .not.quick_method%grad) then
+    if (.not.quick_method%opt) then
         SAFE_CALL(getEnergy(.false.,ierr))
 
         ! One electron properties (ESP, EField)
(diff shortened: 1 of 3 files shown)


src/main.f90
```diff
@@ -214,23 +214,26 @@ program quick
     timer_cumer%TIniGuess=timer_cumer%TIniGuess+timer_end%TIniGuess-timer_begin%TIniGuess &
                           -(timer_end%T2elb-timer_begin%T2elb)
 
-    if (.not.quick_method%opt .and. .not.quick_method%grad) then
+    if (.not.quick_method%opt) then
         SAFE_CALL(getEnergy(.false.,ierr))
 
         ! One electron properties (ESP, EField)
-        call compute_oeprop()
+        if (.not.quick_method%grad) then
```
vtripath65, line 221: I was wondering if we should call compute_oeprop() only once here, instead of both here and where we compute gradients.

vtripath65 (commented):
Looks good. Only a question about how we are calling compute_oeprop.

PR #489, opened 2026-07-10
Tighter cutoff

vtripath65:
Tighten default cutoffs.

1) CUTOFF           1.0d-7       ->         1.0d-10
2) XCCUTOFF      1.0d-7      ->         1.0d-8

- Introduced finer cutoff for basis sets with diffuse basis functions (multiply both CUTOFF and XCCUTOFF by 0.1)
- Primlimit is now made tighter along with integral cutoff in adjust_cutoff
- TIGHTINT and COARSEINT settings are made finer.

PR #492, opened 2026-07-15
Fix keyword detection

vtripath65:
Keywords are now checked for a perfect match. If the user provided keyword is a part of another keyword it is no longer detected.

Added two functions to the quick_input_parser_module.

- index_keyword : Returns the index of the first letter of the keyword
- found_keyword : Returns .true. if the keyword is found.

These function should be used to detect keywords henceforth.


agoetz:
Closing this PR because all changes were included in PR #489

